PR #38, opened 2014-12-15
Allow apply default filters to be passed in as a param

seanculver:
If apply_default_filters is passed as a parameter it is stripped from the params and set as a value in the options.  This allows you to disable the applying of default parameters on any request.

Tried not to modify any existing behavior


cantino:
Updated with master and removed the gemfile change.  Please confirm once more that this is working how you expect @seanculver.

seanculver:
@cantino will do. Thank you for updating master.

Issue #39, opened 2014-12-24
Specs fail on Rails 4.2


seanculver:
One of the issues is that `ActiveRecord::Reflection::ClassMethods::reflections` [has been changed to use string keys rather than symbols](https://github.com/rails/rails/pull/17718), which breaks https://github.com/mavenlink/brainstem/blob/master/lib/brainstem/presenter_collection.rb#L185)  A solution would be to use `reflect_on_association` instead which works with both symbols and strings (and prefers symbols, according to the docs).

```
 association = model.class.reflect_on_association(v.method_name)
```

After the above change is made there are 3 additional spec failures. Further investigation is needed.

cantino:
Will be fixed in #43.

PR #40, opened 2014-12-24
Temporarily patch gemspec to ignore rails 4.2

seanculver:
This is a temporary workaround until #39 is resolved.


cantino:
No longer necessary.

PR #42, opened 2015-01-01
security: sanitize sort direction going into search

cantino:
@rogerneel @ahuth @shirish-pampoorickal appreciate a review.


rogerneel:
LGTM, direction looks sanitized at the source

ahuth:
Specs are failing because of [this](https://github.com/mavenlink/brainstem/blob/b427b38c633f4492a1339367bae028676b4e68b2/lib/brainstem/presenter_collection.rb#L185).

A model's reflections now has string keys, instead of symbols (Rails probably changed on us). We're looking for `reflections[:tasks]` when we should be looking for `reflections["tasks"]`.

I know this isn't big Maven, so it probably doesn't matter, but should we be merging PRs that aren't green on CI / have failing tests? As a reviewer, I can't give a thumbs up until we can see the tests fail correctly, which we can't do here.

cantino:
Just fixed it.  And yes, that was an oversight on my part.

cantino:
I rushed it in a bit since it's security related.

ahuth:
Ok, thanks Andrew. Were we alerted to a security issue here, or how does that work?

cantino:
I found it incidentally, determined it didn't actually affect us, but could affect someone, and fixed it.

PR #46, opened 2015-01-27
Add tooling to help build Brainstem Controllers

cantino:
* Infer the param model name from the controller_name, with optional overload

### DIFF
--- a/lib/brainstem/concerns/controller_param_management.rb
+++ b/lib/brainstem/concerns/controller_param_management.rb
@@ -1,0 +1,20 @@
+module Brainstem
+  module Concerns
+    module ControllerParamManagement
+      extend ActiveSupport::Concern
+
+      included do
+        class_attribute :brainstem_plural_model_name, :brainstem_model_name,
+                        instance_accessor: false, instance_reader: false, instance_writer: false
+      end
+
+      def brainstem_model_name
+        self.class.brainstem_model_name.to_s.presence || controller_name.singularize
+      end
+
+      def brainstem_plural_model_name
+        self.class.brainstem_plural_model_name.to_s.presence || brainstem_model_name.pluralize
+      end
+    end
+  end
+end
--- a/spec/brainstem/concerns/controller_param_management_spec.rb
+++ b/spec/brainstem/concerns/controller_param_management_spec.rb
@@ -1,0 +1,42 @@
+require 'spec_helper'
+require 'brainstem/concerns/controller_param_management'
+
+describe Brainstem::Concerns::ControllerParamManagement do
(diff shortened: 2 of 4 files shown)


lib/brainstem/concerns/controller_param_management.rb
```diff
@@ -0,0 +1,20 @@
+module Brainstem
+  module Concerns
+    module ControllerParamManagement
+      extend ActiveSupport::Concern
+
+      included do
+        class_attribute :brainstem_plural_model_name, :brainstem_model_name,
+                        instance_accessor: false, instance_reader: false, instance_writer: false
```
einzige: having `instance_accessor: false` should be enough

lib/brainstem/concerns/controller_param_management.rb
```diff
@@ -0,0 +1,20 @@
...
+      end
+
+      def brainstem_model_name
+        self.class.brainstem_model_name.to_s.presence || controller_name.singularize
```
einzige: controller name should be singularized by default. Otherwise it wouldn't work in Rails anyway.

spec/brainstem/concerns/controller_param_management_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+require 'brainstem/concerns/controller_param_management'
+
+describe Brainstem::Concerns::ControllerParamManagement do
+  class TasksController
```
einzige: you can apply different approach here. 

```ruby
let :controller_class do
  Class.new do
    include ControllerParamManagement
    def controller_name; 'tasks' end
  end
end

subject(:controller) { controller_class.new }

describe '.brainstem_model_name=' do
  it { expect { controller_class.brainstem_model_name = 'thingy' }.to chnage { controller.brainstem_model_name }.to('thingy') }
end
```

lib/brainstem/concerns/controller_param_management.rb
cantino: Hey @einzige.  No, `controller_name` returns something like `tasks`, not `task`.

lib/brainstem/concerns/controller_param_management.rb
cantino: I thought so too!

But not for Rails 3.  `instance_accessor` was added in Rails 4.

spec/brainstem/concerns/controller_param_management_spec.rb
cantino: Using the anonymous class is a nice trick, thanks!

PR #47, opened 2015-01-30
Adopt rubocop code style

cshaffer:
Greetings! Big fan of brainstem.  This PR introduces [rubocop](https://github.com/bbatsov/rubocop) as a development dependency in hopes of maintaining a consistent code style going forward.

The .rubocop_todo.yml file kept track of all the offenses that rubocop couldn't automatically fix.  If you would like to go this route, I would be willing to go through each pending offense and address accordingly.

No worries if rubocop isn't for you, in which case feel free to close.


cantino:
Hey @cshaffer, thank you!

I'll have to go over some of these changes and think about which way we want to take them.  I agree with many of them, although I don't think the `"` -> `'` is worth it due to the amount of churn in the git blame log that would result.  Also, from my reading, it has no practical speed implications.

cshaffer:
@cantino No problem.  Preferring single quotes when there is no string interpolation needed is purely a style choice and does not provide any runtime benefits.  I can disable that cop, and others if you'd prefer. Let me know.

cshaffer:
I've configured the string literal cop to prefer double quotes. How's this look @cantino?

cantino:
I'm a fan of these changes.  I'm currently working on a new version of Brainstem and would prefer to wait to do this change until after that merge, so that I don't get merge conflicts up the wazoo.

cshaffer:
Sounds like a good plan.  Feel free to ping me after the fact, I'm happy to resolve merge conflicts later, or start over with the fresh code, whatever makes the most sense.

PR #51, opened 2015-04-15
Optimize queries for MySQL

cantino:
In MySQL, selecting only the primary id from an expensive table sort query can sometimes result in faster results.  In this PR, we start always plucking IDs, then fetching the actual results with a second query.


shirish-pampoorickal:
LGTM!

shirish-pampoorickal:
@cantino Can this branch be deleted?

cantino:
Once we have deployed big maven's repointing away from this branch.

---
Andrew Cantino
Chief Architect, Mavenlink

Issue #13, opened 2013-08-15
Conditionally Disable Count from Being Returned


ksylvest:
For large tables including the count can have very large performance implications:

```
SELECT COUNT(id) FROM something; -- Runtime: 5507.556ms
SELECT * FROM something LIMIT 50 OFFSET 0; -- Runtime: 0.193 ms
```

It would be great to be able to send in an option to disable the count entirely.

PR #52, opened 2015-04-19
Adds ability to exclude count from results

ivadnerb:
This relates to issue [13] (https://github.com/mavenlink/brainstem/issues/13). You can exclude counts from results in the following ways:

```ruby
class WidgetPresenter < Brainstem::Presenter
  presents "Widget"

  # 1. exclude the count and do not allow counts to be requested
  exclude_count

  # 2. exclude the count and allow counts to be requested in url, e.g. with "count=true"
  exclude_count allow: true


(etc)

end
```

If the exclude_count method is not used in the presenter definition, the count can be excluded using "count=false" in the url query string.

If you've any questions, just let me know. And by the way, nice gem :)

### DIFF
--- a/lib/brainstem/presenter.rb
+++ b/lib/brainstem/presenter.rb
@@ -28,6 +28,32 @@ module Brainstem
       end
     end
 
+    # @overload exclude_count
+    #   set exclude_count = true and allow_count = false (i.e. exclude the count and do not
+    #   do not allow a request for the count via the url query string)
+    #   @return [Boolean, Boolean]
+    # @overload exclude_count( allow: true )
+    #   set exclude_count = true and allow_count = false (i.e. exclude the count by default
+    #   but allows a request for it via the url query string with (e.g. ?count=true)
+    #   @return [Boolean, Boolean]    
+    # @overload exclude_count( exclude: false )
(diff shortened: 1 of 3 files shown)


cantino:
Hey @ivadnerb, thanks for the contribution!  I'm at RailsConf right now, but I'll dig into this when I get back.

lib/brainstem/presenter.rb
```diff
@@ -28,6 +28,32 @@ def self.default_sort_order(sort_string = nil)
...
+    #   @return [Boolean, Boolean]
+    # @overload exclude_count( allow: true )
+    #   set exclude_count = true and allow_count = false (i.e. exclude the count by default
+    #   but allows a request for it via the url query string with (e.g. ?count=true)
+    #   @return [Boolean, Boolean]    
+    # @overload exclude_count( exclude: false )
+    #   set exclude_count = false. This is only needed for testing to toggle back to the
+    #   including the count by default, otherwise some of the subsequent tests may fail
+    #   @return [Boolean, Boolean]
+    def self.exclude_count(exclude: true, allow: false)
```
cantino: Brainstem still needs to support 1.9, so kwargs can't be used quite yet unfortunately.

PR #54, opened 2015-05-08
Only optimize queries by plucking IDs first for MySQL

cshaffer:
Now that queries are split between plucking IDs and actual records, when paginating and sorting the kind of SQL generated for plucking IDs looks something like this:

```
SELECT  DISTINCT posts.id FROM "posts"  ORDER BY posts.title asc LIMIT 20 OFFSET 0
```
Postgres complains unless you include the column you're ordering by in the select:

```
ActiveRecord::StatementInvalid:
       PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
       LINE 1: SELECT  DISTINCT posts.id FROM "posts"  ORDER BY posts.title asc...
                                                                ^
       : SELECT  DISTINCT posts.id FROM "posts"  ORDER BY posts.title asc LIMIT 20 OFFSET 0
```

This commit adds the calculated sort name to the query when plucking IDs to appease Postgres (or any other database that complains similarly)

### DIFF
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -79,10 +79,14 @@ module Brainstem
 
         # On complex queries, MySQL can sometimes handle 'SELECT id FROM ... ORDER BY ...' much faster than
         # 'SELECT * FROM ...', so we pluck the ids, then find those specific ids in a separate query.
-        ids = scope.pluck("#{scope.table_name}.id")
-        id_lookup = {}
-        ids.each.with_index { |id, index| id_lookup[id] = index }
-        records = scope.klass.where(id: id_lookup.keys).sort_by { |model| id_lookup[model.id] }
+        if(ActiveRecord::Base.connection.instance_values["config"][:adapter] =~ /mysql/)
+          ids = scope.pluck("#{scope.table_name}.id")
+          id_lookup = {}
+          ids.each.with_index { |id, index| id_lookup[id] = index }
+          records = scope.klass.where(id: id_lookup.keys).sort_by { |model| id_lookup[model.id] }
+        else
+          records = scope.to_a
+        end
       end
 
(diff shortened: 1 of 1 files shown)


cantino:
Thanks Chad!

Could we do it for Postgres only?  Or, if you'd like, we could make the two-pass load conditional on MySQL, as it's really an optimization for that DB only.

---
Andrew Cantino
Chief Architect, Mavenlink

cshaffer:
No problem!

If it were up to me, my preference would be to avoid DB specific conditionals and instead prefer a solution that works for anything.

As is, I would think the performance hit to MySQL would be minimal in including this extra column and you'd have much less code to maintain.

Not to mention, unless you plan to go all the way with only on DB and leave users tied to other DB's behind, you're asking for a slippery slope of tricky support for all the DB's and their quirks, on top of dealing with ActiveRecord's quirks with each DB driver gem.

Barring all that, if you'd still prefer switching on DB here, I think it might be less complicated to add my tweak for Postgres only.

Do you have a preferred method for detecting Postgres? A quick google hasn't resulted in anything promising...

cantino:
I agree about avoiding DB conditionals.  The challenge is that I *think* that adding additional columns on the select will reduce the MySQL optimization.  This is unfortunate, and we don't fully understand it, but in MySQL, selecting only the primary key in a complex, table-sorted query is much faster than selecting more columns.  I suspect it has to do with how much data gets copied to the temp table.

In any case, it's silly to bleed this optimization into other DBs.

Something like `ActiveRecord::Base.connection.instance_values["config"][:adapter] == "mysql"` might work?

cshaffer:
Thanks for the help @cantino. Let me know if you'd like anything changed.

cantino:
Looks great, thank you!

PR #56, opened 2015-06-12
PresenterValidator handles more complex preloads; smarter preload de-duping when loading models

cantino:
@robyurkowski mind reviewing / refactoring this a bit?

Hopefully a good introduction to a piece of the new Brainstem.

### DIFF
--- a/lib/brainstem/presenter.rb
+++ b/lib/brainstem/presenter.rb
@@ -185,16 +185,41 @@ module Brainstem
     # Run preloading on the given models.
     def preload_associations!(models, context)
       if models.length > 0
-        association_names_to_preload = context[:requested_associations_hash].values.map(&:method_name).compact
-        association_names_to_preload += configuration[:preloads].to_a
-        # todo: better de-duping here when things might be nested hashes?
-        association_names_to_preload.uniq!
-        if association_names_to_preload.any?
-          reflections = context[:reflections]
-          association_names_to_preload.reject! { |association| !reflections.has_key?(association.to_s) }
-          if association_names_to_preload.any?
-            Brainstem::Presenter.ar_preload(models, association_names_to_preload)
+        reflections = context[:reflections]
+
+        preloads = context[:requested_associations_hash].values.map(&:method_name).compact
+        preloads += configuration[:preloads].to_a
+
+        hash_preloads = {}
+        single_preloads = []
+
+        preloads.each do |preload_name|
+          case preload_name
+            when Hash
+              preload_name.each do |key, value|
+                if hash_preloads[key]
+                  hash_preloads[key.to_s] = Array[hash_preloads[key], value]
+                else
+                  hash_preloads[key.to_s] = value
+                end
+              end
+            when NilClass
+            else
+              single_preloads << preload_name.to_s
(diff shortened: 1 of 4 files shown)


robyurkowski:
:+1: -- will take a look at it on Monday.

lib/brainstem/presenter.rb
```diff
@@ -185,17 +185,42 @@ def calculate_sort_name_and_direction(user_params = {})
...
-        # todo: better de-duping here when things might be nested hashes?
-        association_names_to_preload.uniq!
-        if association_names_to_preload.any?
-          reflections = context[:reflections]
-          association_names_to_preload.reject! { |association| !reflections.has_key?(association.to_s) }
-          if association_names_to_preload.any?
-            Brainstem::Presenter.ar_preload(models, association_names_to_preload)
+        reflections = context[:reflections]
+
+        preloads = context[:requested_associations_hash].values.map(&:method_name).compact
```
robyurkowski: I think the fact that I have to try and parse through the meaning of the context is a good indication we're missing an abstraction here.

lib/brainstem/presenter.rb
```diff
@@ -185,17 +185,42 @@ def calculate_sort_name_and_direction(user_params = {})
...
         end
+
+        single_preloads -= hash_preloads.keys
+
+        single_preloads.uniq!
+
+        single_preloads.reject! { |preload_name| !reflections.has_key?(preload_name) }
+        hash_preloads.reject! { |preload_name, value| !reflections.has_key?(preload_name) }
+
+        single_preloads << hash_preloads if hash_preloads.any?
```
robyurkowski: This is basically data transformation, and it makes a lot of assumptions about what it's expecting in the way of that data without explicitly identifying the API surface / format of the data.

I'm going to read through some more of the source and see if I can't piece together what the collaborating data object looks like. I think at that point it would make a lot of sense to encapsulate the transformation logic into a collaborating object. (Put another way, the presenter shouldn't have to know how to transform/merge/dedupe the list of preloads.)

cantino: I agree.  This is a functional implementation with no attempt yet at extraction.

lib/brainstem/presenter.rb
cantino: Agreed, the `context` is a bag of data that is passed through primarily as an optimization to avoid calling things like `Brainstem::Presenter.reflections(models.first.class),` over and over again.

spec/brainstem/presenter_spec.rb
```diff
@@ -430,6 +430,16 @@ def workspace.synthetic_id
 
         result = presenter.group_present(Workspace.order('id desc'), %w[tasks user lead_user tasks_with_lambda])
       end
+
+      it "includes any nested hash associations declared via the preload DSL directive" do
```
cantino: It'd be fine to make this into a separate test that asserts correct dedeuping across symbols and strings.

robyurkowski:
@cantino As discussed, merge this branch + refactor against primary feature branch?

cantino:
Yup!

PR #57, opened 2015-06-19
Refactor presenter preload

robyurkowski:
I need to add one more fix here before this is merged (missing a test for when there are no valid preloads), but I think this is a reasonable attempt at removing some of the complexity of the presenter into an appropriate collaborator.

### DIFF
--- a/lib/brainstem/preloader.rb
+++ b/lib/brainstem/preloader.rb
@@ -1,0 +1,98 @@
+module Brainstem
+  # Takes a list of arbitrarily nested objects, compacts them, de-duplicates
+  # them, and passes them to ActiveRecord to preload.
+  #
+  # The following is considered a valid data structure:
+  #
+  #     [:workspaces, {"workspaces" => [:projects], "users" }
+  #
+  # Which will produce the following for ActiveRecord:
+  #
+  #     {"workspaces" => [:projects], "users" => []}
+  #
+  class Preloader
+
+    ################################################################################
+    # Class API
+    ################################################################################
+    class << self
+      def preload(*args)
+        new(*args).call
+      end
+    end
+
+    ################################################################################
+    # Instance API
+    ################################################################################
+    attr_accessor :models,
+                  :preloads,
+                  :reflections,
+                  :valid_preloads
+
+    private :valid_preloads=
+
+    attr_writer   :preload_method
+
+    def initialize(models, preloads, reflections, preload_method = nil)
+      self.models         = models
+      self.preloads       = preloads.compact
+      self.reflections    = reflections
+      self.preload_method = preload_method
+      self.valid_preloads = {}
+    end
+
+    def call
+      clean!
+      preload!
+    end
+
+    ################################################################################
+    private
+    ################################################################################
+
+    # De-duplicates, reformats, and prunes requested preloads into an acceptable
+    # format for the preloader
+    def clean!
+      dedupe!
+      remove_unreflected_preloads!
+    end
+
+    def preload!
+      preload_method.call(models, valid_preloads) if valid_preloads.keys.any?
+    end
+
+    # Returns a proc that takes two arguments, +models+ and +association_names+,
+    # which, when called, preloads those.
+    #
+    # @return [Proc] A callable proc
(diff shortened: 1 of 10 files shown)


lib/brainstem/preloader.rb
```diff
@@ -0,0 +1,106 @@
...
+
+
+    def initialize(models, preloads, reflections, preload_method = nil)
+      self.models         = models
+      self.preloads       = preloads.compact
+      self.reflections    = reflections
+      self.preload_method = preload_method
+      self.valid_preloads = {}
+    end
+
```
cantino: This is more whitespace between definitions than we typically use.

lib/brainstem/presenter.rb
```diff
@@ -53,11 +54,6 @@ def self.reflections(klass)
     end
 
     def self.ar_preload(models, association_names)
-      if Gem.loaded_specs['activerecord'].version >= Gem::Version.create('4.1')
```
cantino: why was this removed?  how is it passing without it?

lib/brainstem/preloader.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def preload!
+      preload_method.call(models, valid_preloads) if valid_preloads.keys.any?
+    end
+
+
+    # Returns a proc that takes two arguments, +models+ and +association_names+,
+    # which when called preloads those.
```
cantino: "which, when called, preloads those."

spec/brainstem/preloader_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    context "when preload in the list of reflections" do
+      let(:reflections) { { users: [], posts: [] } }
+
+      it "keeps it" do
+        expect(preloader.valid_preloads.keys).to eq [:users, :posts]
+      end
+    end
+
+    context "when preload not in the list of reflections" do
+      let(:reflections) { { users: [] } }
```
cantino: pretty sure reflections has string keys, not symbols.

spec/brainstem/presenter_spec.rb
```diff
@@ -404,41 +404,73 @@ def workspace.synthetic_id
...
+      #    inappropriately here -- a fact that I think bears weight given
+      #    we're making unit-level assertions in an integration spec.
+      #    However, there's also some value in asserting that these are
+      #    passed through without digging into Rails internals. However,
+      #    further 'purity' improvements could be made by introspecting on
+      #    AR's actual data structures.
+      #
+      #    That's about three shades on the side of overkill, though.
+      #
+      def preloader_should_receive(hsh)
```
cantino: Is all of this to get around stubbing the old `ar_preload`?

lib/brainstem/presenter.rb
robyurkowski: Looks like I forgot to yank the surrounding method somehow. This is injected in Preloader as a proc.

lib/brainstem/preloader.rb
robyurkowski: Tightened that up.

spec/brainstem/presenter_spec.rb
robyurkowski: I've moved `ar_preload` into the Preloader as the presenter shouldn't care about the actual method of preloading. 

Inside `Preloader`, it's appropriate to inject a method or a proc as a collaborator because it should be insulated from the actual implementation within ActiveRecord—or, put another way, what the preloader needs is to send a message to ActiveRecord, but just *which* message is actually an implementation detail of AR and should be injected as a dependency. 

We could move the `ar_preload` method entirely within the Preloader, but that closely binds Preloader to AR. In some situations (testing being the biggest example), we want to inject a different collaborator that handles the way that preloading is done.

robyurkowski:
@cantino I've addressed your feedback here -- if this is good to go, I'll squash before merge.

cantino:
Go ahead and squash.

robyurkowski:
Squashed / rebased.

cantino:
Thanks!

PR #58, opened 2015-08-04
Add the option for filters to be passed the params

naiyt:
@cantino here's the change we discussed. The use case for this is if you have a filter that you want to conditionally change depending on other filters. The use case might look like this:

```ruby
filter :cool_filter, :include_params => true do |scope, opts, params|
  if params[:other_filter].present?
    scope
  else
    scope.cool_scope
  end
end
```

### DIFF
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -225,21 +225,24 @@ module Brainstem
     end
 
     def run_filters(scope, options)
-      extract_filters(options).each do |filter_name, arg|
+      extracted_filters = extract_filters(options, param_opts: true)
+      extracted_filters.each do |filter_name, filter_opts|
+        arg = filter_opts[0]
+        include_params = filter_opts[1]
         next if arg.nil?
         filter_lambda = options[:presenter].filters[filter_name][1]
 
         if filter_lambda
-          scope = filter_lambda.call(scope, arg)
+          scope = include_params ? filter_lambda.call(scope, arg, extracted_filters) : filter_lambda.call(scope, arg)
         else
-          scope = scope.send(filter_name, arg)
+          scope = include_params ? scope.send(filter_name, arg, extracted_filters) : scope.send(filter_name, arg)
         end
       end
 
       scope
     end
 
-    def extract_filters(options)
+    def extract_filters(options, param_opts=false)
       filters_hash = {}
       run_defaults = options.fetch(:apply_default_filters) { true }
 
@@ -250,7 +253,12 @@ module Brainstem
 
         filter_options = filter[0]
         args = run_defaults && requested.nil? ? filter_options[:default] : requested
-        filters_hash[filter_name] = args unless args.nil?
+        include_params = filter_options[:include_params]
+        if param_opts
+          filters_hash[filter_name] = [args, include_params]
+        else
+          filters_hash[filter_name] = args unless args.nil?
+        end
       end
 
       filters_hash
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -225,24 +225,24 @@ module Brainstem
     end
 
     def run_filters(scope, options)
-      extracted_filters = extract_filters(options, param_opts: true)
+      extracted_filters = extract_filters(options)
       extracted_filters.each do |filter_name, filter_opts|
-        arg = filter_opts[0]
-        include_params = filter_opts[1]
-        next if arg.nil?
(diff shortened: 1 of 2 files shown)


lib/brainstem/presenter_collection.rb
```diff
@@ -250,7 +253,12 @@ def extract_filters(options)
 
         filter_options = filter[0]
         args = run_defaults && requested.nil? ? filter_options[:default] : requested
-        filters_hash[filter_name] = args unless args.nil?
+        include_params = filter_options[:include_params]
+        if param_opts
+          filters_hash[filter_name] = [args, include_params]
+        else
+          filters_hash[filter_name] = args unless args.nil?
```
cantino: Maybe make this always point to a hash now like `{ args: args, include_params: include_params }`?

cantino: Then you don't need `param_opts`

lib/brainstem/presenter_collection.rb
```diff
@@ -225,21 +225,24 @@ def handle_only(scope, only)
     end
 
     def run_filters(scope, options)
-      extract_filters(options).each do |filter_name, arg|
+      extracted_filters = extract_filters(options, param_opts: true)
+      extracted_filters.each do |filter_name, filter_opts|
+        arg = filter_opts[0]
```
cantino:
```ruby
extracted_filters.each do |filter_name, filter_details|
  arg = filter_details[:arg]
  include_params = filter_details[:include_params]
```

lib/brainstem/presenter_collection.rb
```diff
@@ -250,7 +253,12 @@ def extract_filters(options)
 
         filter_options = filter[0]
         args = run_defaults && requested.nil? ? filter_options[:default] : requested
```
cantino: Hmm, not your change, but this should probably be called `arg` not `args`.

lib/brainstem/presenter_collection.rb
```diff
@@ -275,7 +278,16 @@ def run_search(scope, includes, sort_name, direction, options)
         search_options[:page] = calculate_page(options)
       end
 
-      search_options.reverse_merge!(extract_filters(options))
+      extracted_filters = extract_filters(options)
```
naiyt: @cantino: I switched to setting the `filters_hash` to `{ arg: arg, include_params: filter_options[:include_params] }`, like you mentioned. The one tricky part is making sure that `run_search` gets the extracted filters in the format it wants, which is what I'm doing here.

lib/brainstem/presenter_collection.rb
```diff
@@ -275,7 +278,16 @@ def run_search(scope, includes, sort_name, direction, options)
...
+      extracted_filters.each do |key, val|
+        if val[:arg].nil?
+          extracted_filters.delete(key)
```
cantino: So search errors if it receives `key => nil`?  What do nil args represent?  Default filters?

Assuming we do need to filter them, I'd probably make a new variable called `extracted_filters_for_search` or something that is key => val[:arg] like

```ruby
extracted_filters.each.with_object({}) { |(key, val), hash| hash[key] = val[:arg] if val[:arg] }
```

naiyt: I'm not entirely sure I understand what they represent, but it does seem to be expected that filters with nil args aren't included in the extracted filters hash here. (`extract_filters` used to do it like this, before these changes: `filters_hash[filter_name] = args unless args.nil?`)

I didn't know that`.each.with_object` trick, that's pretty nifty.

lib/brainstem/presenter_collection.rb
```diff
@@ -275,7 +278,12 @@ def run_search(scope, includes, sort_name, direction, options)
         search_options[:page] = calculate_page(options)
       end
 
-      search_options.reverse_merge!(extract_filters(options))
+      extracted_filters = extract_filters(options)
+      extracted_filters_for_search = extracted_filters.each.with_object({}) do |(key, val), hash|
+        hash[key] = val[:arg] unless val[:arg].nil?
```
cantino: Do we have a spec on this?

naiyt: Yeah, [this spec](https://github.com/mavenlink/brainstem/blob/add_params_to_filters/spec/brainstem/presenter_collection_spec.rb#L619) tests the situation where `val[:arg]` is nil..

cantino:
LGTM!

naiyt:
Cool, thanks for the review. :+1:

PR #59, opened 2015-08-06
Have params hash be key => val instead of key => { arg: val }

naiyt:
@cantino this updates the filters params hash to be `key => val` instead of `key => { arg: val }`.

### DIFF
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -228,7 +228,7 @@ module Brainstem
       extracted_filters = extract_filters(options)
 
       params = extracted_filters.each.with_object({}) do |(key, val), hash|
-        hash[key] = val[:arg]
+        hash[key] = val[:arg] unless val[:arg].nil?
       end
 
       extracted_filters.each do |filter_name, filter_opts|
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -226,6 +226,11 @@ module Brainstem
 
     def run_filters(scope, options)
       extracted_filters = extract_filters(options)
+
+      params = extracted_filters.each.with_object({}) do |(key, val), hash|
+        hash[key] = val[:arg]
+      end
+
       extracted_filters.each do |filter_name, filter_opts|
         arg = filter_opts[:arg]
         include_params = filter_opts[:include_params]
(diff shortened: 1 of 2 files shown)


spec/brainstem/presenter_collection_spec.rb
```diff
@@ -506,13 +506,13 @@ class MyException < Exception; end
       context "with include_params" do
         it "passes the params into the filter block" do
           WorkspacePresenter.filter :filter_with_param, :include_params => true do |scope, option, params|
```
cantino: Should we have a second filter as well and make request that includes both, and show that the one with `include_params` sees the other param?

lib/brainstem/presenter_collection.rb
```diff
@@ -226,6 +226,11 @@ def handle_only(scope, only)
 
     def run_filters(scope, options)
       extracted_filters = extract_filters(options)
+
+      params = extracted_filters.each.with_object({}) do |(key, val), hash|
+        hash[key] = val[:arg]
```
cantino: do we also want `unless val[:arg].nil?`

naiyt: I think there may be situations where you want to see all of the requested filters, even ones without an arg, which is why I wasn't checking for nil there.

spec/brainstem/presenter_collection_spec.rb
naiyt: Good idea, I've added that.

lib/brainstem/presenter_collection.rb
cantino: if they don't have an arg, they're a default filter, right?  can the user send a nil arg?

naiyt: My thought was that you might want to know about the existence of other filters (including default filters), and not just their args. Although I can't think of a specific use case for that. We can check for `val[:arg].nil?` if you think that would never be needed.

cantino: I think it should probably match what we pass to search.

naiyt: Alright, I've removed it

naiyt: Or rather, added the check for nil args.

cantino:
LGTM, thanks!

PR #60, opened 2015-08-07
Compact search results if a value isn't found with the scope.  

bgreg:
It is possible that a search result could have id's that are not found in the scope. We are compacting the search results to protect against
that here.

/cc @cantino @jhmoore


cantino:
Thanks @bgreg, let's see if this fixes it!

PR #61, opened 2015-08-18
Add params to filters

cantino:
@naiyt, how does this look?  In forward-porting your changes to Cerebellum, it seemed like we could simplify by not changing `extract_filters` at all, and instead looking into the configuration when calling a filter to see if it has requested `include_params`.  Do you see any differences between this and master, around how it handles `nil` or default filters, for example?  It seems like it'd be the same.


cantino:
Oops, wrong base.

PR #62, opened 2015-08-18
Add params to filters

cantino:
@naiyt, how does this look?  In forward-porting your changes to Cerebellum, it seemed like we could simplify by not changing `extract_filters` at all, and instead looking into the configuration when calling a filter to see if it has requested `include_params`.  Do you see any differences between this and master, around how it handles `nil` or default filters, for example?  It seems like it'd be the same.

### DIFF
--- a/lib/brainstem/presenter.rb
+++ b/lib/brainstem/presenter.rb
@@ -147,13 +147,17 @@ module Brainstem
     def apply_filters_to_scope(scope, user_params, options)
       helper_instance = fresh_helper_instance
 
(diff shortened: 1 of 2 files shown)


cantino:
Note that I didn't need to do any of the changes to `run_search` as were in https://github.com/mavenlink/brainstem/pull/58.

naiyt:
Nice, that all looks right to me. The handling of `nil` and default filters seems identical to me,  just with the logic for throwing away filters with `nil` args back in `extract_filters`. LGTM, thanks.

PR #63, opened 2015-10-07
use ruby 2.2.3

diclophis:


diclophis:
@ahuth / @jrolfs this should be ok right?

ahuth:
Definitely. LGTM

PR #64, opened 2015-10-30
Add Optional Fields to Presenters

shirish-pampoorickal:
Add the concept of optional fields, requested with an optional_fields param.

### DIFF
--- a/spec/brainstem/dsl/field_spec.rb
+++ b/spec/brainstem/dsl/field_spec.rb
@@ -103,7 +103,7 @@ describe Brainstem::DSL::Field do
   describe "#optioned?" do
     context "when is not an optional field" do
       it 'returns true' do
-        expect(field.optioned?(['expensive_title', 'expensive_title2'])).to eq true
+        expect(field.optioned?(['some_optional_field', 'some_other_optional_field'])).to eq true
       end
     end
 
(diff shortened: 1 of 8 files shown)


shirish-pampoorickal:
@robyurkowski Appreciate a look!  FYI, `name` on `DSL::Field` is now a string to be more consistent with associations.

cc @cantino

cantino:
I'm really pleased with how easy this was now that we've updated the brainstem internals to use a DSL.

---
Andrew Cantino
Chief Architect, Mavenlink

cantino:
@jhmoore, mind reviewing?

cantino:
@jhmoore, mind reviewing?

spec/brainstem/dsl/field_spec.rb
```diff
@@ -99,4 +99,31 @@ def matches?(model, helper_instance, conditional_cache)
       end
     end
   end
+
+  describe "#optioned?" do
+    context "when is not an optional field" do
+      it 'returns true' do
+        expect(field.optioned?(['expensive_title', 'expensive_title2'])).to eq true
+      end
```
jhmoore: Not sure how much we care here, but . . . it might be worth adding an assertion here in a before block of the context that the field we're checking is indeed not an optional field.  It took me several minutes to figure out that `expensive_title` and `expensive_title2` were established as actual fields in a separate spec helper file.

jhmoore: Actually, now that I look at the actual spec helper, it has `optional: true` for both of those.  Is this test returning true because those are optional fields?

cantino: Hey @jhmoore, you're right, that's confusing.  Those values aren't actually being used in the first case because it's not optional, but I updated it to be a little clearer I think.

PR #65, opened 2015-11-05
Add error presentation helpers to Brainstem

cantino:
@shirish-pampoorickal, @plainprogrammer and I extracted these into Brainstem.

/cc @naiyt


naiyt:
Nice, this will definitely be helpful.

PR #66, opened 2015-11-06
Adds api doc generation.

robyurkowski:
At long last, @cantino.

There are still a few outstanding issues that need to be fixed up today, and I'm going to read through all this code again and have a fresh look at it, but this should be workable to look at.

If you're reading through this, I'd strongly recommend you start with `bin/brainstem` and work your way down the stack.

**REMAINING ISSUES**
- [x] Rewire brain to stop using `Cerebellum`
- [x] Arg of ~~`Cerebellum::Presenter`~~ `Brainstem::Presenter#presents` has diverged from `Symbol` to `Class` and this needs to be updated
- [x] Confirm that formatter can handle multiple `presents` declarations
- [x] Ensure configuration flags have `Brainstem::ApiDocs` config options and CLI flag equivalents
- [x] Add recursive check to nested parameter fields ala `WorkspacePresenter`
- [x] Add class-level `brainstem_model_name` method 
- [x] Ensure `restrict_to_only` options documented on Presenter associations
- [x] Ensure 'optional`documentable
- [x] Add `each_filename` to `AbstractCollection`
- [x] Find way to introspect on presenters missing explicitly declared `brainstem_key` field (`brainstem_model_name`)?
- [x] We get an unknown Sink error unless -m is provided.
- [x] Correct builder arity docs
- [x] TODO cleanup

**Remaining**:

**Low-hanging Fruit**
- [x] If no parameters are specified, don't output the "Valid Parameters" section for controllers.
- [x] Demodulize controllers directory and rename to `endpoints`
- [x] Rename models to 'objects' directory
- [x] Add nodoc nonheritable to Presenters 
- [x] Make sure that the controller instance actually responds_to? a method before we output it as an endpoint
- [x] We'd like filters and associations to not output and doc (be nodoc!) unless they have an info: string

**More difficult**:
- [x] Support reloading in development mode (add `clear_configuration!` to controllers that implement config in `Brainstem.reset!`
- [x] Check that conditionals have nodoc option
- [x] Come up with a more manageable way for formatters to reach sibling data
- [x] Associations should have markdown links to the presenter for their associated object (unless it's `:polymorphic`)
- [x] Adding `nodoc!` in a presenter or controller prevents it from being documented, but still outputs an empty file.
- [x] How do we `nodoc: true` a filter, it doesn't seem to be working? (It looks like the only valid options are `default info include_params`.)
- [x] Allow presenter association models to be output / linked even if they are not presented by a controller.
- [ ] Ensure operational documentation complete
- [ ] If no info: is declared on a controller-level valid param, can we use the info from the presenter field of the same name?
- [ ] Maybe we could auto-link exact model name matches in the descriptions, info text?
- [ ] Move `--markdown` to `--format=markdown`
- [ ] Add some tests around Inheritance in controllers

**Where time permits / cleanup**:
- [x] Make sure missing specs are filled in or unnecessary
- [ ] Clean up or remove extraneous / initial sketch sinks
- [ ] Ensure method documentation complete


cantino:
:clap:

lib/brainstem/api_docs.rb
```diff
@@ -0,0 +1,43 @@
...
+module Brainstem
+  module ApiDocs
+    include ActiveSupport::Configurable
+
+    config_accessor(:base_path) do
+      File.expand_path("./brainstem_docs")
+    end
+
+    config_accessor(:controller_filename_pattern) do
+      File.join("controllers", "{{name}}_controller.{{extension}}")
```
cantino: I haven't seen this style before; where does the interpolation on this happen?

lib/brainstem/concerns/formattable.rb
```diff
@@ -0,0 +1,60 @@
+require 'brainstem/api_docs'
+require 'active_support/inflector/inflections'
+
+#
+# TODO: Spec this out
+#
+module Brainstem
+  module Concerns
+    module Formattable
```
cantino: Does this need to be an ActiveSupport::Concern?

robyurkowski: Nothing's done on inclusion or extension, and it doesn't mix class / instance methods, so nope, just a standard module should suffice!

lib/brainstem/api_docs.rb
robyurkowski: This style as in the moustaches? It's used inside of `suggested_filename` methods inside any of the non-collection domain wrappers (`endpoint.rb`, `controller.rb`, `presenter.rb`). It's still a bit kludgy IMO but being able to specify the format of it should add some flexibility — in fact, one thing that might be nice to add would be to add a separate `{{ api_version_path }}` that could be included or removed so that controllers aren't by default output into `/api/v1/`.

robyurkowski: Also, if you have another preferred format, I'd be happy to use it. I debated both `printf` and ERB-style templating for those but this seemed like the most self-explanatory.

lib/brainstem/api_docs/builder.rb
```diff
@@ -0,0 +1,165 @@
...
+
+
+      #
+      # Holds a reference to the constructed introspector.
+      #
+      attr_accessor :introspector
+
+
+      #
+      # A proc of arity 1 which returns a new atlas, which takes info from
```
robyurkowski: This is a lie. It should be a proc of arity `1..2`.

lib/brainstem/api_docs.rb
cantino: No, this is good.

cantino:
I'll let you finish this up and get it passing, then do a close review.

cantino:
Any update on this @robyurkowski?

robyurkowski:
Couple things coming shortly.

*Rob Yurkowski*
Founder, bequeath.al <http://www.bequeath.al>
rob@yurkowski.net | @robyurkowski <https://www.twitter.com/robyurkowski> |
519 . 379 . 3940

robyurkowski:
@cantino Re: `Confirm that formatter can handle multiple presents declarations`, is there any situation where a controller would explicitly present multiple data types per action? I seem to recall — but don't have any notes to this effect — that we'd also like to include the presenter documentation for each of the associations of a specific presenter. Do we want to do this inside the presenter documentation, or do we also want to do this in each action, respectful of the `restrict_to_only` option?

Neither is particularly onerous, so if you'd prefer the latter even though it's more complicated, let me know.

cantino:
If I'm understanding the question correctly, I think the answer is no. I don't think that we have any polymorphic primary returns where we would return a mixture of different types of presenter objects. Brainstem may support that later, but it doesn't yet. Is that what you were asking?

---
Andrew Cantino
Chief Architect, Mavenlink

robyurkowski:
@cantino Exactly, yup. I thought I'd seen `presents` taking multiple classes in one example, but realized later I was looking at a Presenter, which is immaterial to its documentation at the moment.

cantino:
Yes, a presenter can know how to present many classes (usually STI) but a controller will only return one type as the primary scope.

---
Andrew Cantino
Chief Architect, Mavenlink

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - 1.9.3
   - 2.0.0
   - 2.1.1
```
cantino: Let's add 2.2.3 here to.

README.md
```diff
@@ -269,6 +269,85 @@ Brainstem parses the request params and supports the following:
 
 --
 
+## Documenting an API
```
cantino: Maybe "Automatically generating API documentation"?

README.md
```diff
@@ -269,6 +269,85 @@ Brainstem parses the request params and supports the following:
...
+#### Nodoc
+
+The following methods accept a `:nodoc` boolean option, which indicates that the
+documentation should be suppressed for this particular entry:
+
+- `association` &mdash; hides the association
+- `field` &mdash; hides the field
+- `sort_order` &mdash; hides the sort order
+- `filter` &mdash; hides the filter
+- `request` / `model`
```
cantino: "conditionals"

README.md
```diff
@@ -269,6 +269,85 @@ Brainstem parses the request params and supports the following:
...
+
+#### Additional Documentables
+
+In addition to the above, there are three additional methods in the DSL designed
+primarily for documentation:
+
+- `nodoc!` &mdash; skips this presenter entirely and does not document it.
+- `title(str, options)` &mdash; used to specify an alternate title for the
```
cantino: Could you provide an example of this?

README.md
```diff
@@ -269,6 +269,85 @@ Brainstem parses the request params and supports the following:
...
+    Presenter.
+    - `nodoc: true` &mdash; forces fallback to the Presenter's constant
+- `description(str, options)` &mdash; used to specify a description for the
+    Presenter.
+    - `nodoc: true` &mdash; displays no description
+
+
+### Controllers
+
+Fill this in.
```
cantino: ^

README.md
cantino: Also, need documentation on the CLI and generating docs in here.

lib/brainstem/api_docs.rb
```diff
@@ -0,0 +1,99 @@
...
+    #
+    # The following tokens will be substituted:
+    #
+    # - {{name}} : the underscored name of the controller without 'controller'
+    # - {{extension} : the specified file extension
+    #
+    # @see #output_extension
+    #
+    config_accessor(:controller_filename_pattern) do
+      File.join("controllers", "{{name}}_controller.{{extension}}")
```
cantino: Let's change this to just `{{name}}.{{extension}}`.

cantino: In an "endpoints" directory.

lib/brainstem/api_docs.rb
```diff
@@ -0,0 +1,99 @@
...
+    #
+    # The following tokens will be substituted:
+    #
+    # - {{name}} : the demodulized underscored name of the presenter
+    # - {{extension} : the specified file extension
+    #
+    # @see #output_extension
+    #
+    config_accessor(:presenter_filename_pattern) do
+      File.join("models", "{{name}}.{{extension}}")
```
cantino: models -> objects

cantino:
@robyurkowski, you mentioned having a class diagram for these changes?

lib/brainstem/api_docs/builder.rb
```diff
@@ -0,0 +1,165 @@
+require 'brainstem/api_docs/introspectors/rails_introspector'
+require 'brainstem/api_docs/atlas'
+require 'active_support/core_ext/hash/slice'
+
+# TODO: Clean up public/private in this class.
```
cantino: Please do any TODOs or remove them if they're not actually important.

cantino: I should say-- please search this PR for "TODO" and identify any that you feel are important to have in a V1.

lib/brainstem/cli/generate_api_docs_command.rb
```diff
@@ -0,0 +1,161 @@
...
+      #
+      # @return [OptionParser] the option parser that should mutate the
+      #   +options+ hash.
+      #
+      def option_parser
+        OptionParser.new do |opts|
+          opts.banner = "Usage: generate [options]"
+
+          opts.on('-m', '--multifile-presenters-and-controllers',
+                  'dumps presenters and controllers to separate files a la '\
```
cantino: Please drop the Mavenlink reference.

lib/brainstem/cli/generate_api_docs_command.rb
```diff
@@ -0,0 +1,161 @@
...
+                  'controllers. It is matched against the constant, not '\
+                  'underscored name of the controller. Specifying multiple '\
+                  'performs a logical AND between Regexen.') do |o|
+            # Trim slashes on passed MATCH.
+            matcher = Regexp.new(o.gsub(/(\A\/)|(\/\z)/, ''), 'i')
+            options[:builder][:args_for_atlas][:controller_matches].push(matcher)
+          end
+
+
+          opts.on('--markdown', 'use markdown format') do |o|
```
cantino: Be good to also support `--format markdown`?

lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -0,0 +1,306 @@
...
+
+module Brainstem
+  module Concerns
+    module ControllerDSL
+      extend ActiveSupport::Concern
+      include Brainstem::Concerns::InheritableConfiguration
+
+      DEFAULT_BRAINSTEM_PARAMS_CONTEXT = :_default
+
+      included do
```
cantino: We will need to call `clear_configuration!` on controllers that include this, I think, to support Rails class reloading.  Should be added to `Brainstem.reset!`

lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -0,0 +1,306 @@
...
+        # Container method that sets up base scoping for the configuration.
+        #
+        def brainstem_params(&block)
+          self.brainstem_params_context = DEFAULT_BRAINSTEM_PARAMS_CONTEXT
+          class_eval(&block)
+          self.brainstem_params_context = nil
+        end
+
+
+        # Specifies that the scope should should not be documented.
```
cantino: should should

lib/brainstem/concerns/controller_param_management.rb
```diff
@@ -1,22 +1,40 @@
...
-
       def brainstem_model_name
-        self.class.brainstem_model_name.to_s.presence || controller_name.singularize
+        self.class.brainstem_model_name.to_s.presence ||
+          controller_name.singularize
       end
 
       def brainstem_plural_model_name
-        self.class.brainstem_plural_model_name.to_s.presence || brainstem_model_name.pluralize
+        self.class.brainstem_plural_model_name.to_s.presence ||
```
cantino: Why do we need the `||` here if we handle it in the class below?

lib/brainstem/concerns/presenter_dsl.rb
```diff
@@ -100,6 +158,8 @@ def reset_configuration!
           configuration.nest!(:filters)
           configuration.nest!(:sort_orders)
           configuration.nest!(:associations)
```
cantino: Should nodoc be here too?

lib/brainstem/controller_methods.rb.rej
```diff
@@ -0,0 +1,7 @@
+diff a/lib/brainstem/controller_methods.rb b/lib/brainstem/controller_methods.rb	(rejected hunks)
```
cantino: Don't think you meant to commit this.

lib/brainstem/dsl/configuration.rb
```diff
@@ -11,8 +11,13 @@ class Configuration
...
       #   which the new configuration object should use as a base.
       def initialize(parent_configuration = nil)
-        @parent_configuration = parent_configuration || ActiveSupport::HashWithIndifferentAccess.new
-        @storage = ActiveSupport::HashWithIndifferentAccess.new
+        @parent_configuration  = parent_configuration || ActiveSupport::HashWithIndifferentAccess.new
+        @storage               = ActiveSupport::HashWithIndifferentAccess.new
+
+        self.nonheritable_keys =
+          parent_configuration.respond_to?(:nonheritable_keys) &&
+          parent_configuration.nonheritable_keys ||
```
cantino: So we *inherit* our parent's non-inheritable key list?  If the parent then changes this list, do we change?

lib/brainstem/presenter.rb
```diff
@@ -136,7 +137,7 @@ def extract_filters(user_params, options = {})
         user_value = user_value.is_a?(Array) ? user_value : (user_value.present? ? user_value.to_s : nil)
         user_value = user_value == "true" ? true : (user_value == "false" ? false : user_value)
 
-        filter_options = filter[0]
+        filter_options = filter
```
cantino: Could probably remove this variable now.

lib/brainstem/presenter.rb
```diff
@@ -150,10 +151,11 @@ def apply_filters_to_scope(scope, user_params, options)
 
       requested_filters = extract_filters(user_params, options)
       requested_filters.each do |filter_name, filter_arg|
-        filter_lambda = configuration[:filters][filter_name][1]
+        filter_lambda = configuration[:filters][filter_name][:value]
 
         args_for_filter_lambda = [filter_arg]
-        args_for_filter_lambda << requested_filters if configuration[:filters][filter_name][0][:include_params]
+        args_for_filter_lambda << requested_filters \
```
cantino: Rather not to a post-if on the next line.  If you want to wrap, let's not use a post-if.

lib/brainstem/presenter.rb
```diff
@@ -168,7 +170,7 @@ def apply_filters_to_scope(scope, user_params, options)
     # Given user params, apply a validated sort order to the given scope.
     def apply_ordering_to_scope(scope, user_params)
       sort_name, direction = calculate_sort_name_and_direction(user_params)
-      order = configuration[:sort_orders][sort_name]
+      order = configuration[:sort_orders].fetch(sort_name, {})[:value]
```
cantino: Why was the fetch needed here?

cantino: I guess it's effectively just a `try`?

spec/brainstem/api_docs/abstract_collection_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+require 'brainstem/api_docs/abstract_collection'
+
+module Brainstem
+  module ApiDocs
+    describe AbstractCollection do
+      let(:member) { Object.new }
+
+
+      describe "#last" do
+        it "retrieves the last member"
```
cantino: Missing spec

lib/brainstem/concerns/controller_param_management.rb
robyurkowski: Good catch.

lib/brainstem/concerns/presenter_dsl.rb
robyurkowski: I think it should be for consistency. If you can you think of any reason why it would be good to have this remain heritable, I'll change it back, but until then I'll add it.

lib/brainstem/presenter.rb
robyurkowski: Effectively, yep, though I don't particularly remember making this choice as a preference over introducing a dependency on ActiveSupport given it's peppered in throughout. However, doing it this way keeps our spec suite just a little bit leaner, and I don't think it's particularly cryptic... If you'd rather I make a change to `configuration[:sort_orders].try(:[], sort_name).try(:[], :value)`, I can.

lib/brainstem/dsl/configuration.rb
robyurkowski: I feel like I maybe don't quite understand the question here.

If the parent changes this list, child classes would also reflect that change. But unless I'm grossly misunderstanding the way that configuration is likely to be used, an annotation of non-heritability is more than likely to be limited to this library's code, rather than in any end-user code which might lead to a breakdown in functionality. As of right now, the use of `nonheritable!` is constrained to the `title`, `description`, and `nodoc!` annotations, all of which (to my mind) should be *inheritably* non-heritable.

Basically, if we suddenly decide that `:associations` is non-heritable, or decide that `:title` should be heritable, we're going to have to make changes, but the trade-off to this is that we don't constantly have to be re-adding these annotations to sub-configurations, especially in the action context where we're manually instantiating `Configuration` objects, and where the non-heritability of `title` still remains important.

But then again, I could very well be misunderstanding the question. `Configuration` so easily twists my head about. It's the Inception of data structures.

cantino: That makes sense.  My question was about if this works:

```ruby
parent = Parent.new
child = Child.new
child.nonheritable!(:a)
parent.nonheritable!(:b)
```

what does `child.nonheritable_keys` print now?  With the array handling in the configuration, it would be an append only set and would include both, but your code might only have the child's?

lib/brainstem/concerns/controller_dsl.rb
robyurkowski: Took a closer look at this, and as we discussed, we are okay here.

shirish-pampoorickal:
Issues:
- [ ] Adding `nodoc!` in a presenter or controller prevents it from being documented, but still outputs an empty file.
- [ ] How do we `nodoc: true` a filter, it doesn't seem to be working? (It looks like the only valid options are `default info include_params`.)

lib/brainstem/dsl/configuration.rb
shirish-pampoorickal: Did you check this?

robyurkowski: So I just checked this. It works as expected in that direction... but there is also something unexpected to me here:

```ruby
[1] pry(main)> parent = Brainstem::DSL::Configuration.new
=> #<Brainstem::DSL::Configuration:0x007fb25a91ee98 @nonheritable_keys=[], @parent_configuration={}, @storage={}>
[2] pry(main)> child = Brainstem::DSL::Configuration.new(parent)
=> #<Brainstem::DSL::Configuration:0x007fb259b4d170
 @nonheritable_keys=[],
 @parent_configuration=#<Brainstem::DSL::Configuration:0x007fb25a91ee98 @nonheritable_keys=[], @parent_configuration={}, @storage={}>,
 @storage={}>
[3] pry(main)> child.nonheritable!(:a)
=> ["a"]
[5] pry(main)> parent.nonheritable!(:b)
=> ["a", "b"]
[6] pry(main)> parent.nonheritable!
ArgumentError: wrong number of arguments (0 for 1)
from /Users/mavenlink/workspace/brainstem/lib/brainstem/dsl/configuration.rb:67:in `nonheritable!'
[7] pry(main)> parent
=> #<Brainstem::DSL::Configuration:0x007fb25a91ee98 @nonheritable_keys=["a", "b"], @parent_configuration={}, @storage={}>
[8] pry(main)> child
=> #<Brainstem::DSL::Configuration:0x007fb259b4d170
 @nonheritable_keys=["a", "b"],
 @parent_configuration=#<Brainstem::DSL::Configuration:0x007fb25a91ee98 @nonheritable_keys=["a", "b"], @parent_configuration={}, @storage={}>,
 @storage={}>
[9] pry(main)> child.nonheritable_keys
=> ["a", "b"]
```

Look at `[5]`. That seems like counter-intuitive and incorrect behaviour to me (where the child also affects the parent). Does that seem wrong to you, or is that an intentional feature?

robyurkowski: Okay, so I've spent a bit more time on this. There is a bug here but I also think it's important that this be documented somewhere, so that future developers can understand the rationale behind programming the nonheritable keys this way. (I've also included it inside the configuration documentation.)

```ruby
        #
        # Nonheritable keys are a bit peculiar: they make the lookup for a key
        # specified as nonheritable to return no result when it falls back to
        # the parent configuration.
        #
        # These keys themselves are inheritable; in this way, a class that
        # descends from another will keep the same behaviour as its superclass
        # without necessarily having the same data. Or to put it another way,
        # if you have specified that 'title' is not inheritable in a
        # superclass's configuration, that is a property of that class, and
        # descendent classes should behave the same way.
        #
        # It is also unlikely that subclasses will modify the list of
        # nonheritable keys.
```

Here is the output with the fix:

```ruby
[1] pry(main)> parent = Brainstem::DSL::Configuration.new
=> #<Brainstem::DSL::Configuration:0x007fade41ffa10 @nonheritable_keys=#<Brainstem::DSL::Configuration::InheritableAppendSet:0x007fade41ff920 @parent_array=[], @storage=[]>, @parent_configuration={}, @storage={}>
[2] pry(main)> child = Brainstem::DSL::Configuration.new(parent)
=> #<Brainstem::DSL::Configuration:0x007fade41853f0
 @nonheritable_keys=#<Brainstem::DSL::Configuration::InheritableAppendSet:0x007fade4185378 @parent_array=#<Brainstem::DSL::Configuration::InheritableAppendSet:0x007fade41ff920 @parent_array=[], @storage=[]>, @storage=[]>,
 @parent_configuration=#<Brainstem::DSL::Configuration:0x007fade41ffa10 @nonheritable_keys=#<Brainstem::DSL::Configuration::InheritableAppendSet:0x007fade41ff920 @parent_array=[], @storage=[]>, @parent_configuration={}, @storage={}>,
 @storage={}>
[3] pry(main)> parent.nonheritable!(:a)
=> ["a"]
[4] pry(main)> child.nonheritable_keys
=> #<Brainstem::DSL::Configuration::InheritableAppendSet:0x007fade4185378 @parent_array=#<Brainstem::DSL::Configuration::InheritableAppendSet:0x007fade41ff920 @parent_array=[], @storage=["a"]>, @storage=[]>
[5] pry(main)> child.nonheritable_keys.to_a
=> ["a"]
[6] pry(main)> child.nonheritable!(:b)
=> ["b"]
[7] pry(main)> parent.nonheritable_keys
=> #<Brainstem::DSL::Configuration::InheritableAppendSet:0x007fade41ff920 @parent_array=[], @storage=["a"]>
[8] pry(main)> parent.nonheritable_keys.to_a
=> ["a"]
[9] pry(main)> child.nonheritable_keys.to_a
=> ["a", "b"]
```

cantino: That does seem wrong

---
Andrew Cantino
Chief Architect, Mavenlink

robyurkowski: Fixed now :)

cantino: Thanks for digging in

---
Andrew Cantino
Chief Architect, Mavenlink

lib/brainstem/presenter.rb
```diff
@@ -24,6 +26,20 @@ def self.presents(*klasses)
...
 
+
+    #
+    # Returns the set of possible brainstem keys for the classes presented.
+    #
+    # If the presenter specifies a key, that will be returned as the only
+    # member of the set.
+    #
+    def self.possible_brainstem_keys
+      @possible_brainstem_keys || begin
```
cantino: What sets `@possible_brainstem_keys`?

robyurkowski: Typo — see bab4303 :)

lib/brainstem/api_docs/endpoint.rb
```diff
@@ -0,0 +1,234 @@
...
+      def to_s
+        "#{http_methods.join(" / ")} #{path}"
+      end
+
+
+      #
+      # Merges http methods (for de-duping Rails' routes).
+      #
+      def merge_http_methods!(methods)
+        self.http_methods |= methods
```
ahuth: Bitwise or assignment?

rogerneel:
LGTM! At long last.

cantino:
Woot! 🎈 🎈 🎆

PR #69, opened 2015-12-05
Allow configuration of links independent from filenames

shirish-pampoorickal:
@robyurkowski, we wanted to be able to have the generated links skip the extension (since we put them in individual directories with middleman for pretty routes), but still want extensions on the actual filenames.  How's this look?

### DIFF
--- a/lib/brainstem/api_docs.rb
+++ b/lib/brainstem/api_docs.rb
@@ -40,6 +40,36 @@ module Brainstem
 
 
     #
+    # Defines the naming pattern for the relative link to each controller documentation file.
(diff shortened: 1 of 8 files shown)


robyurkowski:
@shirish-pampoorickal This looks quite reasonable to me. The one thing I'm curious whether Middleman rewrites links during the build phase if `activate :directory_indexes` has been declared. Has this been checked with an actual `middleman build`? If not, I'm happy to check it out.

cantino:
This did work for us locally with our middleman setup.

PR #71, opened 2015-12-10
Only render the Data Model section when one has been provided

shirish-pampoorickal:
@robyurkowski, this looks better to us.


robyurkowski:
Also LGTM!

PR #72, opened 2015-12-10
Validate that the brainstem_key is required on presenters that accept multiple classes

shirish-pampoorickal:
@robyurkowski, this helps get around the issue with many possible `brainstem_key`s.

### DIFF
--- a/lib/brainstem/presenter.rb
+++ b/lib/brainstem/presenter.rb
@@ -18,7 +18,7 @@ module Brainstem
         if klasses.any? { |klass| klass.is_a?(String) || klass.is_a?(Symbol) }
           raise "Brainstem Presenter#presents now expects a Class instead of a class name"
         end
-        @presents += klasses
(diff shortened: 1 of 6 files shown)


lib/brainstem/presenter.rb
```diff
@@ -18,7 +18,7 @@ def self.presents(*klasses)
         if klasses.any? { |klass| klass.is_a?(String) || klass.is_a?(Symbol) }
           raise "Brainstem Presenter#presents now expects a Class instead of a class name"
         end
-        @presents += klasses
+        @presents.concat(klasses).uniq!
```
robyurkowski: We could use a `Set` here, alternatively.

robyurkowski:
LGTM!

lib/brainstem/presenter.rb
shirish-pampoorickal: True, but likely not worth it-- it's changed infrequently and only on load.

PR #74, opened 2015-12-17
Shift the default sort order to the top and otherwise make alphabetical

shirish-pampoorickal:
@robyurkowski Appreciate a look


robyurkowski:
Seems reasonable to me!

PR #76, opened 2016-04-19
Filter and search query strategy

naiyt:
This PR adds the ability to utilize your Brainstem filters in conjunction with the results returned by your search implementation.

To enable this for a presenter, you just add:

```ruby
query_strategy :filter_and_search
```

Otherwise Brainstem will default to its normal behavior (ignoring all filters if you attempt to search).

This works by calling your search block first, doing `scope.where(id: search_ids)`, then applying the filters, order, and pagination  on the resultant scope.

It currently sends a limit of 10,000 to your search block for performance reasons. It's necessary to retrieve all possible ids from your search because in order for pagination to work properly we have to know the total number of possible matches. If we don't retrieve all possible ids from the search implementation it's impossible to know what the total count of possible matches is (which is the intersection of filtered and searched results). This could be potentially slow, so any implementation should be thoroughly performance tested (it appears to be relatively fast with around ~10,000 entities when tested against ElasticSearch, and much faster when there are a smaller number of results).

If you need your search to return more than 10,000 results then this solution will not work.

[This README has some more details.](https://github.com/mavenlink/brainstem/blob/filter_and_search/lib/brainstem/query_strategies/README.md)

@shirish-pampoorickal @jhmoore + whoever else, what do you guys think of this?

cc @ahuth @arleigh-atkinson @eihli @lambdatastic @jasonwc @pyluftig

### DIFF
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -1,6 +1,6 @@
 require 'brainstem/search_unavailable_error'
 require 'brainstem/presenter_validator'
-require_relative './pagination_strategies/filter_or_search'
+require_relative './query_strategies/filter_or_search'
 
 module Brainstem
   class PresenterCollection
@@ -50,7 +50,7 @@ module Brainstem
       options[:default_per_page] = default_per_page
       options[:default_max_per_page] = default_max_per_page
 
-      strategy = Brainstem::PaginationStrategies::FilterOrSearch.new(options)
+      strategy = Brainstem::QueryStrategies::FilterOrSearch.new(options)
       primary_models, count = strategy.execute(scope)
 
       # Determine if an exception should be raised on an empty result set.
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -53,6 +53,7 @@ module Brainstem
       scope = options[:primary_presenter].apply_filters_to_scope(scope, options[:params], options)
 
       scope, ordered_search_ids = search(options, scope) if searching? options
+      scope_ids = []
 
       if options[:params][:only].present?
         # Handle Only
@@ -79,9 +80,10 @@ module Brainstem
 
           _, next_ordered_ids = search(recursive_paginate_options, scope_before_search)
 
-          scope_ids = scope.pluck(:id)
+          scope_ids = scope_ids + scope.pluck(:id)
+          ordered_search_ids = ordered_search_ids + next_ordered_ids
 
-          scope = scope_before_search.where(id: (scope_ids + next_ordered_ids).uniq)
+          scope = scope_before_search.where(id: ordered_search_ids & scope_ids)
         end
       end
 
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -153,7 +153,7 @@ module Brainstem
               end
 
       return Brainstem::QueryStrategies::FilterOrSearch.new(options) if strat == :legacy
-      return Brainstem::QueryStrategies::FilterAndSearch.new(options) if strat == :filter_and_search && searching?(options) && scope.count <= default_max_filter_and_search_page
+      return Brainstem::QueryStrategies::FilterAndSearch.new(options) if strat == :filter_and_search && searching?(options)
       return Brainstem::QueryStrategies::FilterOrSearch.new(options)
     end
 
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -146,13 +146,8 @@ module Brainstem
     private
 
     def strategy(options, scope)
-      strat = if options[:primary_presenter].configuration.has_key? :query_strategy
-                options[:primary_presenter].configuration[:query_strategy]
-              else
-                :legacy
-              end
+      strat = options[:primary_presenter].get_query_strategy
 
-      return Brainstem::QueryStrategies::FilterOrSearch.new(options) if strat == :legacy
       return Brainstem::QueryStrategies::FilterAndSearch.new(options) if strat == :filter_and_search && searching?(options)
       return Brainstem::QueryStrategies::FilterOrSearch.new(options)
     end
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -18,7 +18,7 @@ module Brainstem
     def initialize
       @default_per_page = 20
       @default_max_per_page = 200
-      @default_max_filter_and_search_page = 70_000 # TODO: figure out a better default and make it configurable
+      @default_max_filter_and_search_page = 10_000 # TODO: figure out a better default and make it configurable
     end
 
     # The main presentation method, converting a model name and an optional scope into a hash structure, ready to be converted into JSON.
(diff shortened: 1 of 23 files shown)


ahuth:
Nicely done, @naiyt. Particularly the :cheese:

lib/brainstem/query_strategies/README.md
```diff
@@ -0,0 +1,94 @@
...
+ActiveRecord filtering provided by Brainstem.
+
+You must take the following important notes into account when using the `filter_and_search` query strategy:
+
+- Your search block should behave the same as it always has: it should return an array where the first element is an array
+  of model ids and the second element is the total number of matched records.
+- This works by retrieving all possible ids from your search block (up to 10,000) and then applying your filters
+  on those returned ids. This has some obvious potential performance considerations as you are potentially returning
+  and querying off of 10,000 results.
+- If you have less than 10,000 possible results you shouldn't have to worry about ordering, because the order will
```
cantino: I recall running some tests like `where(id: [1...20,000])` ids and seeing the MySQL was slow.  You should check and see what that performance looks like for you.

lib/brainstem/presenter_collection.rb
```diff
@@ -175,34 +145,20 @@ def validate!
...
+                options[:primary_presenter].configuration[:query_strategy]
+              else
+                :legacy
+              end
 
-    def calculate_per_page(options)
-      per_page = [(options[:params][:per_page] || options[:per_page] || default_per_page).to_i, (options[:max_per_page] || default_max_per_page).to_i].min
-      per_page = default_per_page if per_page < 1
-      per_page
+      return Brainstem::QueryStrategies::FilterOrSearch.new(options) if strat == :legacy
```
cantino: Seems like you could delete this line?

cantino:
Hello @naiyt! :)

naiyt:
Oh hey, a wild @cantino appears! :D How ya been?

cantino:
Good, how are you? :)

lib/brainstem/query_strategies/README.md
naiyt: We've tested it with 10,000 ids and it's been fast. I think it's most likely quick because while we're still giving it 10,000 potential ids it's still getting paginated with a `limit` / `offset`. Seems MySQL knows how to optimize that?

Our main bottleneck in that scenario is Elasticsearch. We've tested against 100,000 possible objects (which is limited to a total of 10,000 when searching with this) and performance is decent (but not super great). We're planning on doing a fair bit more performance testing to see how feasible this is.

Sadly, we couldn't think up any other way of doing it, since we have to know the total count of possible ids for pagination to work. =\

lib/brainstem/presenter_collection.rb
naiyt: Ah, true!

lib/brainstem/query_strategies/README.md
cantino: Yea, it's a hard problem.  It'd be awesome if this is fast enough!

naiyt: When I try this with a search that matches 10,000+ entities I'm getting an http request of around 1.5-2.5 seconds (much faster when the search returns less results - around 300ms or so). I think it could be optimized a bit as well, potentially.

naiyt:
Doing good! Thanks for dropping by w/some comments.

naiyt:
Hey @bennapp would you be able to give this a review when you get a chance?

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+      end
+
+      def paginate(scope)
+        if @options[:params][:limit].present? && @options[:params][:offset].present?
+          limit = calculate_limit
+          offset = calculate_offset
+        else
+          limit = calculate_per_page
+          offset = limit * (calculate_page - 1)
```
bennapp: Why is it `(calculate_page - 1)`?

lib/brainstem/query_strategies/filter_or_search.rb
```diff
@@ -0,0 +1,113 @@
...
+          offset = calculate_offset
+        else
+          limit = calculate_per_page
+          offset = limit * (calculate_page - 1)
+        end
+
+        [scope.limit(limit).offset(offset).uniq, scope.select("distinct #{scope.connection.quote_table_name @options[:table_name]}.id").count]
+      end
+
+      def handle_only(scope, only)
```
bennapp: Do we need to do anything special for `only` in `FilterAndSearch`?

lib/brainstem/query_strategies/filter_and_search.rb
naiyt: This was mostly copy-pasta from the original pagination method, but I believe the reason is so that in your url params you can say `page=1` for the first page, but the final offset in the sql query properly zero indexes that.

lib/brainstem/query_strategies/filter_or_search.rb
naiyt: My idea is that `only` should be ignored with searching, because it feels like different features. Is there a use case where you would want to search off only specific ids? (The original implementation was similar, in that `only` is ignored when searching.)

bennapp: I didn't have anything in mind, I just wanted to make sure we had all of our bases covered.

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -0,0 +1,49 @@
...
+    class FilterAndSearch < BaseStrategy
+      def execute(scope)
+        ordered_search_ids = run_search(scope, filter_includes.map(&:name))
+        scope = scope.where(id: ordered_search_ids)
+        scope = @options[:primary_presenter].apply_filters_to_scope(scope, @options[:params], @options)
+        scope = @options[:primary_presenter].apply_ordering_to_scope(scope, @options[:params])
+        count = scope.count
+        scope = paginate(scope)
+
+        [scope, count]
```
bennapp: Is the scope the actual models at this point?

naiyt: At this point I believe it's still the arel scope.

lib/brainstem/presenter_collection.rb
```diff
@@ -46,50 +49,17 @@ def presenting(name, options = {}, &block)
...
-        if(ActiveRecord::Base.connection.instance_values["config"][:adapter] =~ /mysql|sqlite/i)
-          ids = scope.pluck("#{scope.table_name}.id")
-          id_lookup = {}
-          ids.each.with_index { |id, index| id_lookup[id] = index }
-          primary_models = scope.klass.where(id: id_lookup.keys).sort_by { |model| id_lookup[model.id] }
-        else
-          primary_models = scope.to_a
-        end
-      end
+      primary_models, count = strategy(options, scope).execute(scope)
```
bennapp: I am a little confused that in `FilterAndSearch` we return an arel scope, but here it would appear we want the instantiated models.

bennapp:
@naiyt and I agreed we should add the SQL optimization in `FilterOrSearch` into `FilterAndSearch`. We are also going to make sure `execute` of `FilterAndSearch` instantiates the models like `execute` of `FilterOrSearch`.

naiyt:
Hey @bennapp, I added an `execute_scope` method to the `BaseStrategy` like we talked about. Now both strategies should evaluate their scopes before returning from the `execute` method.

bennapp:
@naiyt Awesome Job!!! LGTM!

naiyt:
Thanks @bennapp! Performance has been good from testing (and this can be easily be disabled by just changing the query strategy) so I'm going to merge this.

ahuth:
:metal: Well done, Nate

PR #77, opened 2016-04-29
Generate a lookup for complicated fields and associations available to dynamic lambdas

bennapp:
This PR will allow presenters to define `custom_lookup(models, fields)` which should return a 'lookup' hash containing the included fields as keys with their values as a hash with model ids as keys and the complicated/costly fields and/or associations as values. This will allow presenters to do complicated work for their complicated fields before presenting each individual model and then pass a reference to the values for those fields when presenting.

eg.
```
  lookup 
  => { 'some_field' => {'1': 'some value', '2' => 'another value' }, 'another_field' => {...} }
```

### Before
```
      fields do
        field :complicated_field, :string,
              dynamic: lambda { |model| model.complicated_query(current_user) } # This is an N + 1
        #...
      end
```

### After

```
      def custom_lookup(models, fields)
        lookup = {}
        
        if fields.include?(:complicated_field)
          lookup[:complicated_field] = Hash[complicated_query_for_models(models, current_user).select(:id, :value)]
        end

        lookup
      end

      fields do
        field :complicated_field, :string,
              dynamic: lambda { |model, lookup| lookup[:complicated_field][model.id] }
        #...
      end
```

TODO: Specify an option for fields / associations, eg. `use_lookup: true`, which would be equivalent to having `dynamic: lambda { |model, lookup| lookup[:complicated_field][model.id] }` so we don't have to write this same lambda for fields / associations which use the lookup in this simple case.

### DIFF
--- a/lib/brainstem/dsl/association.rb
+++ b/lib/brainstem/dsl/association.rb
@@ -18,8 +18,18 @@ module Brainstem
         end
       end
 
-      def run_on(model, helper_instance = Object.new)
-        options[:dynamic] ? helper_instance.instance_exec(model, &options[:dynamic]) : model.send(method_name)
+      def run_on(model, lookup = {}, helper_instance = Object.new)
+        proc = options[:dynamic]
+
+        if proc
+          if proc.arity > 1
+            helper_instance.instance_exec(model, lookup, &proc)
+          else
+            helper_instance.instance_exec(model, &proc)
+          end
+        else
+          model.send(method_name)
+        end
       end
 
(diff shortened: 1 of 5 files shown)


bennapp:
cc @shirish-pampoorickal

bennapp:
I am closing this PR and opening a new one now that I have a better idea of how to do this.

bennapp:
I opened the new PR here: https://github.com/mavenlink/brainstem/pull/77

PR #78, opened 2016-05-03
Add support for hash filter values. 

shirish-pampoorickal:
When a hash is passed in as a filter value, it gets striingified as shown below:
```
  presenter.extract_filters({ 'owned_by' => { :ids => [2], 2 => [1] }}

  returns "{ 'owned_by' => { :ids => [2], 2 => [1] }"
```

This PR fixes the issue with passing in a hash value for a filter & returns a hash rather a stringified version of the hash

### DIFF
--- a/lib/brainstem/presenter.rb
+++ b/lib/brainstem/presenter.rb
@@ -132,7 +132,7 @@ module Brainstem
       apply_default_filters = options.fetch(:apply_default_filters) { true }
 
       configuration[:filters].each do |filter_name, filter|
-        user_value = sanitize_filter_value(user_params[filter_name])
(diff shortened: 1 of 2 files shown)


shirish-pampoorickal:
@naganowl / @jrolfs Appreciate a look

juanca:
LGTM! Thanks @shirish-pampoorickal!

jrolfs:
Looks great, thanks @shirish-pampoorickal!

cantino:
Just be sure that whatever method you're passing user data to sanitizes it and doesn't assume it's a string / integer. ;)

shirish-pampoorickal:
@cantino 👍

PR #79, opened 2016-05-03
Add a lookup cache option for associations and fields

bennapp:
This PR allows fields and associations to define a lookup option in order to generate a lookup cache which will run once for all the models presented. When a model is presented, who's field/association defines a lookup option, it will refer to the cache to find its value rather than redoing the work.

### Before
```
      fields do
        field :complicated_field, :string,
              dynamic: lambda { |model| model.complicated_query(current_user) } # This is an N + 1
        #...
      end
```

### After
```
      fields do
        field :complicated_field, :string,
              # This should return a hash with the keys as model ids and their values as the values for this field
              lookup: lambda { |models| ComplicatedQueryServiceObject(current_user, models).group_by { |model| model.id } } 
        #...
      end
```

TODO:
- [x] Add specs for Associations
- [x] Add specs for Fields
- [x] Implement lookup support for Fields
- [x] Add specs and implementation for `lookup_fetch` for Fields
- [x] Add specs and implementation for `lookup_fetch` for Associations
- [x] Add README docs with example of `lookup` and `lookup_fetch`

### DIFF
--- a/README.md
+++ b/README.md
@@ -515,7 +515,7 @@ Brainstem provides a rich DSL for building presenters.  This section details the
   an optional documentation string, and a number of options. By default, fields will call a model method with the same
   name as the field's name and return the result. Use the `:via` option to call a different method, or the `:dynamic` option
   to provide a lambda that takes the model and returns the field's output value. Fields which result in N + 1 queries can be
-  optimized with a `:lookup` option, detailed in the `lookup` secontion below. Fields can be conditionally returned with the
+  optimized with a `:lookup` option, detailed in the `lookup` section below. Fields can be conditionally returned with the
   `:if` option, detailed in the `conditionals` section below.  Expensive fields can be declared as `optional: true` so that they are
    only returned when `optional_fields=field` is provided in the API request. Here are some example fields:
 
@@ -571,7 +571,9 @@ presenting and every presented model gets its assocation or value from the cache
 `lookup` lambda takes in the presented models and should generate a cache containing the models' coresponding assocations
 or values. Brainstem expects the return result of the `lookup` to be a Hash where the keys are the presented models' ids
 and the values are those models' associations or values. Use the `lookup` when you would like to preload but cannot
-e.g. if your association references `current_user`.
+e.g. if your association references `current_user`. If the `lookup` opition is defined, the `dynamic` option is defined and
+one model is being presented, then the `dynamic` will be used. If multiple models are being presented and both options are defined,
+the `lookup` will be used.
 
   ```ruby
   associations do
--- a/README.md
+++ b/README.md
@@ -514,7 +514,8 @@ Brainstem provides a rich DSL for building presenters.  This section details the
   Fields have a name, which is what they will be called in the returned JSON, a type which is used for API documentation,
   an optional documentation string, and a number of options. By default, fields will call a model method with the same
   name as the field's name and return the result. Use the `:via` option to call a different method, or the `:dynamic` option
-  to provide a lambda that takes the model and returns the field's output value. Fields can be conditionally returned with the
+  to provide a lambda that takes the model and returns the field's output value. Fields which result in N + 1 queries can be
+  optimized with a `:lookup` option, detailed in the `lookup` secontion below. Fields can be conditionally returned with the
   `:if` option, detailed in the `conditionals` section below.  Expensive fields can be declared as `optional: true` so that they are
    only returned when `optional_fields=field` is provided in the API request. Here are some example fields:
 
@@ -540,6 +541,7 @@ Brainstem provides a rich DSL for building presenters.  This section details the
   class, an optional documentation string, and some options. By default, associations will call the association or
   method on the model with their name. Like fields, you can use `:via` to call a different method or association and
   `:dynamic` to provide a lambda that takes the model and returns a model, array of models, or relation of models.
+  Associations which result in N + 1 queries can be optimized with a `:lookup` option, detailed in the `lookup` secontion below.
 
   If you have an association that tends to be large and expensive to return, you can annotate it with the
   `restrict_to_only: true` option and it will only be returned when the `only` URL param is provided and contains a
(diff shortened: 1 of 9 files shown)


lib/brainstem/dsl/association.rb
```diff
@@ -18,8 +18,22 @@ def method_name
...
+
+        if dynamic_proc
+          if dynamic_proc.arity > 0
+            helper_instance.instance_exec(model, &dynamic_proc)
+          else
+            helper_instance.instance_exec(&dynamic_proc)
+          end
+        elsif lookup_proc
+          context[:lookup][:associations][name] ||= helper_instance.instance_exec(context[:models], &lookup_proc)
+          context[:lookup][:associations][name][model.id]
```
bennapp: Raise if lookup does not respond to `[]`. Check if there is a `lookup_fetch` proc.

bennapp:
cc. @naiyt

bennapp:
@naiyt I would appreciate a review here. I am finished with the implementation. Now I am just following up with the README.

shirish-pampoorickal:
@bennapp Are we assuming that we are always going to use a lookup for the field. What happens when I need to fetch data for a single object & lookup is kind of overkill. Should we have a fallback `dynamic` proc which it calls if the lookup is not specified?

bennapp:
@shirish-pampoorickal Yes, when fetching a single model the lookup will be slightly overkill, however I think the tradeoff is worth it. I think a nice follow up would be: when one model is presented it uses a dynamic the if it is defined.

shirish-pampoorickal:
@bennapp What do you think about leveraging the `restrict_to_only` option for this?

bennapp:
@shirish-pampoorickal I am not entirely sure what you have in mind. It will be incredibly useful to have this option available for associations which do not have `restrict_to_only` set.

lib/brainstem/dsl/association.rb
```diff
@@ -47,6 +40,27 @@ def polymorphic?
...
       end
+
+      private
+
+      def run_on_with_lookup(model, context, helper_instance)
+        context[:lookup][:associations][name] ||= begin
+          proc = options[:lookup]
+          lookup = helper_instance.instance_exec(context[:models], &proc)
+          if !options[:lookup_fetch].present? && !lookup.respond_to?(:[])
+            raise(StandardError, 'The returned result of the lookup lambda must respond to [] since the default `lookup_fetch` relys on the `[] method` in order to access the model\'s assocation(s) from the lookup. Default: lookup_fetch: lambda { |lookup, model| lookup[:assoication_name][model.id] }`')
```
bennapp: spelling relies

README.md
```diff
@@ -563,6 +565,44 @@ Brainstem provides a rich DSL for building presenters.  This section details the
     association :associated_objects, :polymorphic, "a mixture of objects related to this Widget"
   end
   ```
+  
+* `lookup` - Use this option to avoid N + 1 queries for Fields and Associations. The `lookup` lambda runs once when
+presenting and every presented model gets its assocation or value from the cache the `lookup` lambda generates. The
+`lookup` lambda takes in the presented models and should generate a cache containing the models' coresponding assocations
+or values. Brainstem expects the return result of the `lookup` to be a Hash where the keys are the presented models' ids
+and the values are those models' associations or values. Use the `lookup` when you would like to preload but cannot
+e.g. if your association references `current_user`. If the `lookup` opition is defined, the `dynamic` option is defined and
```
bennapp: `opition `

README.md
```diff
@@ -563,6 +565,44 @@ Brainstem provides a rich DSL for building presenters.  This section details the
     association :associated_objects, :polymorphic, "a mixture of objects related to this Widget"
   end
   ```
+  
+* `lookup` - Use this option to avoid N + 1 queries for Fields and Associations. The `lookup` lambda runs once when
+presenting and every presented model gets its assocation or value from the cache the `lookup` lambda generates. The
+`lookup` lambda takes in the presented models and should generate a cache containing the models' coresponding assocations
+or values. Brainstem expects the return result of the `lookup` to be a Hash where the keys are the presented models' ids
```
bennapp: Use this in error: `Brainstem expects the return result of the `lookup` to be a Hash`

README.md
bennapp: `the `dynamic` option is defined,`

README.md
```diff
@@ -563,6 +565,44 @@ Brainstem provides a rich DSL for building presenters.  This section details the
...
+    association :current_user_groups, Group, "the Groups for the current user",
+                lookup: lambda { |models| 
+                  Group.where(subject_id: models.map(&:id)
+                    .where(user_id: current_user.id)
+                    .group_by { |group| group.subject_id } 
+                }
+  end
+	```
+* `lookup_fetch` - Use this option for Fields and Associations if you would like to override how a model should retrieve
+ its value or assocation from the cache returned by the `lookup` cache. The `lookup_fetch` lambda takes in the return
```
bennapp: assocation returned by the `lookup` cache.

bennapp: The `lookup_fetch` lambda takes in the presented model and result from the `lookup` lambda

README.md
```diff
@@ -563,6 +565,44 @@ Brainstem provides a rich DSL for building presenters.  This section details the
...
+ result from the `lookup` lambda and the presented `model`. It should return the `model`'s assocation or value. If
```
bennapp: It should return the association or value from the lookup cache for that model.

README.md
```diff
@@ -563,6 +565,44 @@ Brainstem provides a rich DSL for building presenters.  This section details the
...
+ `lookup_fetch` is not defined, Brainstem will run the default. The example `lookup_fetch` below is equivalent to the default.
+
+  ```ruby
+  fields do
+    field :current_user_post_count, Post, "the number of Posts the current_user has made for this model",
+                lookup: lambda { |models| 
+                  lookup = Post.where(subject_id: models.map(&:id)
```
bennapp: Make this less indented.

README.md
```diff
@@ -563,6 +565,44 @@ Brainstem provides a rich DSL for building presenters.  This section details the
...
+                    .where(user_id: current_user.id)
+                    .group_by { |post| post.subject_id } 
+                  
+                  lookup
+                },
+                lookup_fetch { |lookup, model| lookup[model.id] }
```
bennapp: `lookup_fetch: lambda { |lookup, model| lookup[model.id] }`

bennapp:
cc. @dezman and @naomiajacobs

lib/brainstem/dsl/association.rb
```diff
@@ -18,8 +18,19 @@ def method_name
         end
       end
 
-      def run_on(model, helper_instance = Object.new)
-        options[:dynamic] ? helper_instance.instance_exec(model, &options[:dynamic]) : model.send(method_name)
+      def run_on(model, context, helper_instance = Object.new)
+        if options[:dynamic] && (!options[:lookup] || context[:models].size == 1)
+          proc = options[:dynamic]
+          if proc.arity > 0
```
naiyt: Minor: would this be more explicit if it was `proc.arity == 1`? Since it looks like this only ever supports one arg?

lib/brainstem/dsl/association.rb
```diff
@@ -29,6 +40,27 @@ def polymorphic?
...
       end
+
+      private
+
+      def run_on_with_lookup(model, context, helper_instance)
+        context[:lookup][:associations][name] ||= begin
+          proc = options[:lookup]
+          lookup = helper_instance.instance_exec(context[:models], &proc)
+          if !options[:lookup_fetch].present? && !lookup.respond_to?(:[])
+            raise(StandardError, 'The returned result of the lookup lambda must respond to [] since the default `lookup_fetch` relys on the `[] method` in order to access the model\'s assocation(s). Default: lookup_fetch: lambda { |lookup, model| lookup[:assoication_name][model.id] }`')
```
naiyt: Small typo in `relys` (Should be "relies").

lib/brainstem/dsl/field.rb
```diff
@@ -51,6 +53,27 @@ def conditionals_match?(model, presenter_conditionals, helper_instance = Object.
           presenter_conditionals[conditional].matches?(model, helper_instance, conditional_cache)
         }
       end
+
+      private
+
+      def run_on_with_lookup(model, context, helper_instance)
```
naiyt: This seems like it may be identical to the `Association#run_on_with_lookup` - think it'd be worth added it as a mixin or something?

naiyt:
@bennapp this looks pretty sweet. To make sure I'm fully understanding:

- Right now in Brainstem if I have a complex association the necessary query gets run for every model.
- This allows you to create a lookup table that only gets run once and cached off. The table is structured as `model id => [associated objects]`.
- The default fetch just grabs the associated objects keyed off of by the model id
- You can define a custom fetch if the structure of your lookup is different in some way

Is that all correct? Does it seem to work when tested manually?

lib/brainstem/dsl/field.rb
bennapp: This is addressed here: https://github.com/mavenlink/brainstem/pull/79/commits/712b392d86e2958ce8cc63e519a945f97fe6959d

bennapp:
@naiyt all of that is correct. I have a working example here: https://github.com/mavenlink/mavenlink/pull/5629

lib/brainstem/dsl/field.rb
naiyt: Nice 👍

bennapp:
@naiyt @shirish-pampoorickal Could I get a final review here?

naiyt:
Sorry for not getting back to you earlier @bennapp, I was out sick on Friday. This LGTM

shirish-pampoorickal:
@bennapp This LGTM. Like the flexibility with `dynamic` & `lookup_fetch`! I would cut a branch of our APP & test if all the specs are green

PR #80, opened 2016-05-17
Avoid pagination in search implementation in favor of pagination in arel

JoeGaebel:
This allows for appropriate server count when developers are paginating their search implementations. 

Currently just throwing this out there to gauge the correctness of this approach. Seems like using any type of search with pagination will only return a fixed sized array of result ids. In order to fix this, we realized that pagination can happen in the arel layer while the full set of results can be enumerated by the search implementation (for example, elastic search).

This is analogous to searching with filters, where the search implementation returns a huge result set and leaves filtering and pagination in the arel layer.

Currently, the specs were not touched so there should be a few failing. Depending on people's interests, @juanca is willing to fix and add more spec coverage. 

-- @juanca & @joegaebel

@naiyt (but limited to) Whattar your thoughts?

### DIFF
--- a/lib/brainstem/query_strategies/filter_or_search.rb
+++ b/lib/brainstem/query_strategies/filter_or_search.rb
@@ -49,19 +49,20 @@ module Brainstem
           order: { sort_order: sort_name, direction: direction },
         )
 
-        if @options[:params][:limit].present? && @options[:params][:offset].present?
-          search_options[:limit] = calculate_limit
-          search_options[:offset] = calculate_offset
-        else
-          search_options[:per_page] = calculate_per_page
-          search_options[:page] = calculate_page
-        end
-
         search_options.reverse_merge!(@options[:primary_presenter].extract_filters(@options[:params], @options))
 
-        result_ids, count = @options[:primary_presenter].run_search(@options[:params][:search], search_options)
+        result_ids, _ = @options[:primary_presenter].run_search(@options[:params][:search], search_options)
+
+        if @options[:params][:limit].present? && @options[:params][:offset].present?
+          limit = calculate_limit
+          offset = calculate_offset
+        else
+          limit = calculate_per_page
+          offset = limit * (calculate_page - 1)
+        end
+
         if result_ids
-          [scope.where(id: result_ids), count, result_ids]
+          [scope.where(id: result_ids).limit(limit).offset(offset).uniq, result_ids.length, result_ids]
         else
           raise(SearchUnavailableError, 'Search is currently unavailable')
         end


lib/brainstem/query_strategies/filter_or_search.rb
```diff
@@ -49,19 +49,20 @@ def run_search(scope, includes, sort_name, direction)
...
+          limit = calculate_per_page
+          offset = limit * (calculate_page - 1)
         end
 
-        search_options.reverse_merge!(@options[:primary_presenter].extract_filters(@options[:params], @options))
-
-        result_ids, count = @options[:primary_presenter].run_search(@options[:params][:search], search_options)
         if result_ids
-          [scope.where(id: result_ids), count, result_ids]
+          [scope.where(id: result_ids).limit(limit).offset(offset).uniq, result_ids.length, result_ids]
```
naiyt: You might be able to use the existing `paginate` method for this thing.

naiyt:
So currently, do you not get a proper total count when using search? Is there not a way for ElasticSearch to tell you the total number of possible results?

lib/brainstem/query_strategies/filter_or_search.rb
```diff
@@ -49,19 +49,20 @@ def run_search(scope, includes, sort_name, direction)
           order: { sort_order: sort_name, direction: direction },
         )
 
+        search_options.reverse_merge!(@options[:primary_presenter].extract_filters(@options[:params], @options))
```
naiyt: This would probably be a good situation to create a new `query_strategy`, because this possibly changes how your presenter and search block need to interact with Brainstem. It would probably be better to keep this working as is for legacy reasons.

juanca:
I'm not sure the result set from tire has the full count. I'll explore that data structure a bit more and follow up.

Though, the current implementation in `brainstem/master` is broken since it would use the `result_ids.length` instead of another property on the `result_ids`. I know we're using elastic search in Mavenlink but, as an open source library, this should do the right thing disregarding the search implementation. 

Perhaps a new query strategy might be the easiest way to introduce these changes though I'd like the default strategy to be more robust.

Also, I think filtering and searching query strategy might be the ideal default strategy for future usages of brainstem (though, I'm also happy with leaving as a different non-default strat)

naiyt:
Yeah, it would be good for the default strategy to be more robust, I just don't think we should introduce breaking changes to it if we can help it. If someone had hacked in their own pagination stuff into their search block I wonder if this could potentially cause some conflicts.

I'd rather not have filter and search be the default, due to all the caveats that come with it -- you have to ensure that your search and your query maintain the same order (which is real difficult if you want complex Brainstem ordering), your search implementation needs to be able to support really big page sizes, and it doesn't support greater than 10,000 possible results.

juanca:
> I just don't think we should introduce breaking changes to it if we can help it

This shouldn't break anything since all it changes is how it determines real server count (actual `result_ids.length`) and paginates with SQL (arel `limit` and `offset` so data is not wastefully loaded). 

> If someone had hacked in their own pagination stuff into their search block I wonder if this could potentially cause some conflicts.

Yeah, I noticed some implementations that limited their search query to certain `per_page` and `page` parameters. In that case, this should return the same response since the server count will be the same (`result_ids.length`) and it will just get the same data from SQL.

> I'd rather not have filter and search be the default, due to all the caveats that come with it -- you have to ensure that your search and your query maintain the same order (which is real difficult if you want complex Brainstem ordering), your search implementation needs to be able to support really big page sizes, and it doesn't support greater than 10,000 possible results.

Interesting. I agree it'd be better to be more conservative. I guess this is for another conversation elsewhere buuut: I was under the impression that filter_and_search could do arbitrary lengths of data since it's a simple `scope.where(id: result_ids).limit(...).offset(...).order(...)` allowing for the same type of ordering as a non-search query. Seems like I read that wrong.

naiyt:
> This shouldn't break anything since all it changes is how it determines real server count (actual result_ids.length) and paginates with SQL (arel limit and offset so data is not wastefully loaded).

Gotcha, cool.

> Yeah, I noticed some implementations that limited their search query to certain per_page and page parameters. In that case, this should return the same response since the server count will be the same (result_ids.length) and it will just get the same data from SQL.

Yeah,  makes sense. I'm also thinking that since this is an open source project we want to make sure that we either communicate clearly what changes like this entail, or introduce some sort of versioning to Brainstem, along with changelogs and stuff. i.e., I think we'd be okay at Mavenlink but it would be good to make it clear for others using Brainstem what updating to the latest version might entail.

> Interesting. I agree it'd be better to be more conservative. I guess this is for another conversation elsewhere buuut: I was under the impression that filter_and_search could do arbitrary lengths of data since it's a simple scope.where(id: result_ids).limit(...).offset(...).order(...) allowing for the same type of ordering as a non-search query. Seems like I read that wrong.

It can't support arbitrary lengths because in order for pagination to work at all we have to request the ENTIRE set of results from search. Otherwise there's no way to properly take the intersection and get the total server count. That's why we limited it to 10,000: requesting page sizes of 10,000 from ElasticSearch is still pretty performant, and gives us enough results for most use cases.

cantino:
Does Results::Collection respond to `total_pages` or `total_entries`?

juanca:
I'll check out the methods on `Results::Collection` @cantino 

Also, I noticed that `run_search` returns a second variable which looks like it could be the server count (which could just be `total_pages`/`total_entries`. I'll double check that

JoeGaebel:
We utilized the `total` method which returns the total count

juanca:
Closing this since the real solution was to return the real count via the search implementation: `result_ids, total_count`

PR #82, opened 2016-09-16
Fix duplicates when paginating over records sorted by non-unique values

jhmoore:
Right now, as an example, if you have many records with identical `updated_at` timestamps, and you sort by `updated_at`, the resulting sort will be non-deterministic, and records will randomly show up multiple times on different pages.

We are adding the primary key (descending) as a fallback tie-breaker sort to _all_ queries -- one possible worry here is that it could adversely affect performance, but we think that most databases will do the right thing if given a query of `order by id desc, id desc`.

### DIFF
--- a/spec/brainstem/presenter_spec.rb
+++ b/spec/brainstem/presenter_spec.rb
@@ -715,7 +715,8 @@ describe Brainstem::Presenter do
         end
 
         sql = presenter.apply_ordering_to_scope(scope, 'order' => 'title').to_sql
-        expect(sql).to match(/order by workspaces.title desc, workspaces.id desc/i)
+        expect(sql).to match(/order by workspaces.title desc, workspaces.id desc, workspaces.id asc/i)
+        # this should be ok, since the first id sort will never have a tie
       end
 
       it 'chains the primary key onto the end' do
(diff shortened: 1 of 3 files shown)


jhmoore:
cc @juanca / @racerpeter

spec/brainstem/presenter_spec.rb
```diff
@@ -680,37 +680,80 @@ def preloader_should_receive(hsh)
...
+        expect(direction).to eq 'asc'
+      end
+
+      it 'can chain multiple sorts together' do
+        presenter_class.sort_order(:title) do |scope|
+          scope.order('workspaces.title desc').order('workspaces.id desc')
+        end
+
+        sql = presenter.apply_ordering_to_scope(scope, 'order' => 'title').to_sql
+        expect(sql).to match(/order by workspaces.title desc, workspaces.id desc/i)
```
shirish-pampoorickal: How is the order for `workspaces.id` still `desc`? We clearly append the order('workspaces.id asc') at the end.

spec/brainstem/presenter_spec.rb
```diff
@@ -680,37 +680,80 @@ def preloader_should_receive(hsh)
...
-        some_method
-        direction = d
-        scope
+    context 'when the sort is not a proc' do
+      it 'applies the named ordering in the given direction' do
+        presenter_class.sort_order :title, 'workspaces.title'
+        expect(presenter.apply_ordering_to_scope(scope, 'order' => 'title:asc').to_sql).to match(/order by workspaces.title asc/i)
+      end
+
+      describe 'deterministic ordering' do
```
shirish-pampoorickal: `deterministic ordering` is not an optional thing, so do we need a nested context for this? Shouldn't there be just one spec

spec/brainstem/presenter_spec.rb
jhmoore: Haha yeah, the order here is actually `order by workspaces.title desc, workspaces.id desc, workspaces.id asc` -- but the match is only checking the first two.

spec/brainstem/presenter_spec.rb
jhmoore: Fair point!

spec/brainstem/presenter_spec.rb
shirish-pampoorickal: This is confusing. Is it worth doing a match for /{table_name}.#{primary_key}/ & then conditionally adding the deterministic order?

jhmoore: We tried; it got kind of crazy examining the arel meta-objects that get created when you pass in symbols.  Also if you pass in a sort of just `id` without the table name, and some other weirdo edge cases.

shirish-pampoorickal: I guess the `deterministic ordering` will never come into play, as the there will never be the conflict on ordering of the primary keys if present

shirish-pampoorickal:
giphy

rogerneel:
@cantino thought you'd be interested in this too

cantino:
Hey Mavens!

This makes sense, and I think it should work. You could argue that users of Brainstem should provide deterministic ordering procs in their Presenters instead of Brainstem always having to append the primary key, but I don't feel strongly.  I would suggest using `scope.connection.quote_table_name` and `scope.connection.quote_column_name` since someone could have a terribly named table that has a space or period in it.  You also should ignore tables that don't have a primary_key, it's not required AFAIK.

jhmoore:
@cantino I like it.  Those also might actually make it easier to test the Arel case, as well, since I believe it also auto-inserts quotes.

Haha yeah -- we thought about what if somebody has a table with multiple primary keys as well, but ActiveRecord seems to explode in rather dramatic fashion if you try.

jhmoore:
@juanca what do you think?

jhmoore:
Travis builds on Ruby versions < 2.2 are still failing because of a Bundler bug

juanca:
Still lgtm

PR #83, opened 2016-09-16
Travis ruby versions

juanca:
@jhmoore @shirish-pampoorickal @racerpeter 

Seems like activesupport will try to install version 5+ and it needs ruby 2+


juanca:
Okay, I just locked it down to the latest of previous major version for ActiveSupport and RSpec. Life is good.

jhmoore:
👍  thanks!

PR #85, opened 2016-12-02
Refactor lookup to take precendece over dynamic lambda

bennapp:
@plainprogrammer ,

This PR is what we had outlined from our previous discussion [here](https://github.com/mavenlink/mavenlink/pull/6878#issuecomment-263328677). Now, if we define a `lookup` that will be ran instead of the `dynamic`. This way we do not run into the case where we are improving the performance of a field or association, adding a `lookup`, and then not having the previous specs cover the new `lookup`.

As a follow up, it might make sense to delete the old dynamics, thoughts?

I made a mavenlink PR just, just to see if we get a passing build on Circle CI 
https://github.com/mavenlink/mavenlink/pull/7141

cc. @shirish-pampoorickal

### DIFF
--- a/spec/brainstem/dsl/association_spec.rb
+++ b/spec/brainstem/dsl/association_spec.rb
@@ -94,24 +94,11 @@ describe Brainstem::DSL::Association do
           }
         }
 
-        context 'when presenting one model' do
-          let(:models) { [first_model] }
-
-          it 'calls the dynamic lambda and not the lookup lambda' do
-            instance = Object.new
-            mock(instance).dynamic_instance_method
-            mock(instance).lookup_instance_method.never
-            expect(association.run_on(first_model, context, instance)).to eq 'Ben'
-          end
-        end
-
-        context 'when presenting more than one model' do
-          it 'calls the lookup lambda and not the dynamic lambda' do
-            instance = Object.new
-            mock(instance).lookup_instance_method
-            mock(instance).dynamic_instance_method.never
-            expect(association.run_on(first_model, context, instance)).to eq 'Ben'
-          end
+        it 'calls the dynamic lambda and not the lookup lambda' do
+          instance = Object.new
+          mock(instance).dynamic_instance_method.never
+          mock(instance).lookup_instance_method
+          expect(association.run_on(first_model, context, instance)).to eq 'Ben'
         end
       end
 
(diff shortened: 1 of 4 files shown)


bennapp:
Also, this is green locally. I am not sure what is up with Travis.

bennapp:
@shirish-pampoorickal or @plainprogrammer I would appreciate a review here.

spec/brainstem/dsl/association_spec.rb
```diff
@@ -94,24 +94,11 @@
...
-        end
-
-        context 'when presenting more than one model' do
-          it 'calls the lookup lambda and not the dynamic lambda' do
-            instance = Object.new
-            mock(instance).lookup_instance_method
-            mock(instance).dynamic_instance_method.never
-            expect(association.run_on(first_model, context, instance)).to eq 'Ben'
-          end
+        it 'calls the dynamic lambda and not the lookup lambda' do
```
shirish-pampoorickal: Shouldn't this say `calls the lookup lambda and not the dynamic lambda`?

shirish-pampoorickal: Should we also rename the context to `with lookup & dynamic lambda`

spec/brainstem/dsl/field_spec.rb
```diff
@@ -102,24 +102,11 @@
...
-        end
-
-        context 'when presenting more than one model' do
-          it 'calls the lookup lambda and not the dynamic lambda' do
-            instance = Object.new
-            mock(instance).lookup_instance_method
-            mock(instance).dynamic_instance_method.never
-            expect(field.run_on(first_model, context, instance)).to eq "Ben's Project"
-          end
+        it 'does not use the dynamic lambda' do
```
shirish-pampoorickal: Same here. Should the context be called `with lookup & dynamic lambda`?

bennapp: Ya, my bad. Thanks.

spec/brainstem/dsl/association_spec.rb
bennapp: Yes, it should say `calls the lookup lambda and not the dynamic lambda`.

bennapp:
Specs are green locally, merging!

PR #87, opened 2017-02-10
Update RSpec to version 3.5 & Patch RR

shirish-pampoorickal:
- Update RSpec to 3.5
- Add RR patch to make it work with the RSpec version
- Get rid of deprecation warnings

- Another PR https://github.com/mavenlink/brainstem/pull/88 has been created to make brainstem compatible with Ruby 1.9.3 & greater

PR #88, opened 2017-02-10
Downgrade ActiveSupport to enable compatibility with Ruby version >= 1.9.3 

shirish-pampoorickal:
- This PR also includes commits from https://github.com/mavenlink/brainstem/pull/87 which ideally should be merged first
- The downgrade caused a single failing spec which was due to an irrelevant extra space in the query. This was fixed by adding the squish method.
screen shot 2017-02-10 at 7 28 57 am


cantino:
Ruby 1.9.3 is no longer supported, why do we need to allow it? Or am I misunderstanding? I believe our usage of ActiveSupport isn't 3.2 compatible. It had a different API for some things. Maybe [this](https://github.com/mavenlink/brainstem/commit/f151c0c12d8b873cdcb7e912120bd7db034f59b5) was one of them?

jhmoore:
Hmm, interesting -- do we run specs in 1.9.3?

shirish-pampoorickal:
@jhmoore Yes, we run specs for ruby 1.9.3, 2.0.0, 2.1.1 & 2.2.3. @cantino There was a commit https://github.com/mavenlink/brainstem/commit/bd8210637fa8e79226f612863460ee3592797db4 specifically added  to support 1.9.3 in Sept 2016, so I assumed we wanted to support it. I would be happy to drop support for 1.9.3 but the specs are still failing for 2.0.0 and 2.1.1, will try to fix those atleast.

@juanca Any reason to support Ruby 1.9.3 anymore?

rogerneel:
I don't see a reason to continue 1.9.3 support. It was EOL'd in Feb 2015.

juanca:
[#83 was the PR for that particular commit](https://github.com/mavenlink/brainstem/pull/83).

Looks like the intent was to get green builds. And it was possible to support 1.9.3 so it left things unchanged (in case someone was using 1.9.3). 

Feel free to drop support for it.

shirish-pampoorickal:
Replaced by https://github.com/mavenlink/brainstem/pull/89

PR #89, opened 2017-02-17
Adjust supported Ruby versions

ahuth:
Replaces #87 and #88.

Drops support for unsupported Ruby versions and adds support for the current stable Ruby releases. Additionally, keeps Ruby 2.2.3, even though it's not a current stable release, because Mavenlink is on it.

cc @shirish-pampoorickal


spec/brainstem/presenter_spec.rb
```diff
@@ -776,7 +776,7 @@ class Cthulhu < Workspace
       end
 
       it 'does not add a fallback deterministic sort, and you deserve whatever fate befalls you' do
-        sql = presenter.apply_ordering_to_scope(scope, order).to_sql
+        sql = presenter.apply_ordering_to_scope(scope, order).to_sql.squish
```
ahuth: Ruby 2.1.10 has an extra space before the `ORDER BY`.

ahuth:
Good idea. Implemented https://github.com/mavenlink/mavenlink/pull/7537 and all tests are green in https://circleci.com/gh/mavenlink/mavenlink/56683?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link.

Thanks Andrew!

PR #92, opened 2017-04-04
Deprecate description param as a string. Going forward the `info` option will be used for the description

shirish-pampoorickal:
#### This branch does the following:
- Allows specifying the description for conditionals, fields and associations with the info option
- Support legacy description param as a string
- Throw a deprecation warning when description param is specified as a string

#### Supported legacy format of conditionals, fields and association:
```
conditionals do
  model   :title_is_hello, lambda { workspace.title == 'hello' }, 'visible when the title is hello'
  request :user_is_bob, lambda { current_user.username == 'bob' }, 'visible only to bob'
end
```
#### Expected format going forward:
```
conditionals do
  model   :title_is_hello, lambda { workspace.title == 'hello' }, info: 'visible when the title is hello'
  request :user_is_bob, lambda { current_user.username == 'bob' }, info: 'visible only to bob'
end
```

### DIFF
--- a/lib/brainstem/dsl/base_block.rb
+++ b/lib/brainstem/dsl/base_block.rb
@@ -27,10 +27,18 @@ module Brainstem
           if maybe_description.is_a?(Hash)
             options = options.merge(maybe_description)
           elsif maybe_description.present?            
+            deprecated_description_warning
             options[:info] = maybe_description
           end
 
           options
+        end
+
+        def deprecated_description_warning
+          ActiveSupport::Deprecation.warn(
+           'DEPRECATION_WARNING: Specifying description as the last parameter will be deprecated in the next version.' \
+           'Description can be specified with the `info` key in a hash. e.g. { info: "My description" }'
+          )
         end
       end
     end
--- a/lib/brainstem/dsl/base_block.rb
+++ b/lib/brainstem/dsl/base_block.rb
@@ -26,7 +26,7 @@ module Brainstem
 
           if maybe_description.is_a?(Hash)
             options = options.merge(maybe_description)
-          elsif maybe_description.present?            
+          elsif maybe_description.present?
             deprecated_description_warning
             options[:info] = maybe_description
           end
@@ -37,7 +37,8 @@ module Brainstem
         def deprecated_description_warning
           ActiveSupport::Deprecation.warn(
            'DEPRECATION_WARNING: Specifying description as the last parameter will be deprecated in the next version.' \
-           'Description can be specified with the `info` key in a hash. e.g. { info: "My description" }'
+           'Description can be specified with the `info` key in a hash. e.g. { info: "My description" }',
+           caller
           )
         end
       end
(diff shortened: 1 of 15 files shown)


lib/brainstem/dsl/base_block.rb
```diff
@@ -22,8 +22,23 @@ def descend(klass, new_config = configuration, new_options = {}, &block)
 
         def parse_args(args)
           options = args.last.is_a?(Hash) ? args.pop : {}
-          description = args.shift
-          [description, options]
+          maybe_description = args.shift
+
+          if maybe_description.is_a?(Hash)
+            options = options.merge(maybe_description)
```
plainprogrammer: What is the case where this would expect to receive two hashes for `args`?

lib/brainstem/dsl/base_block.rb
```diff
@@ -22,8 +22,23 @@ def descend(klass, new_config = configuration, new_options = {}, &block)
 
         def parse_args(args)
           options = args.last.is_a?(Hash) ? args.pop : {}
-          description = args.shift
-          [description, options]
+          maybe_description = args.shift
+
+          if maybe_description.is_a?(Hash)
```
plainprogrammer: Avoiding `is_a?` could be nice — `kind_of?` would be a nominal improvement, but if we could depend on how these objects quack it would be better still.

juanca:
Do we want to deprecate immediately or provide both APIs for the time being?

shirish-pampoorickal:
@juanca This was an effort to let developers know that we are going to deprecated the description as string param in the next version of Brainstem

juanca:
Oh. For some odd reason I was only seeing the new code without any support for description. Whoops, I see the whole now.

lib/brainstem/dsl/base_block.rb
ahuth: I think it's if using the old (non hash) syntax. In that case, `options` will be an empty Hash (see line 24 of this file.

lib/brainstem/dsl/base_block.rb
cantino: You could add a spec showing it works with a HashWithIndifferentAccess too.

cantino:
Makes sense to me. Not sure why I started this with a string description instead of an option. :)

lib/brainstem/dsl/base_block.rb
```diff
@@ -22,8 +22,24 @@ def descend(klass, new_config = configuration, new_options = {}, &block)
...
+            options[:info] = maybe_description
+          end
+
+          options
+        end
+
+        def deprecated_description_warning
+          ActiveSupport::Deprecation.warn(
+           'DEPRECATION_WARNING: Specifying description as the last parameter will be deprecated in the next version.' \
+           'Description can be specified with the `info` key in a hash. e.g. { info: "My description" }',
```
juanca: Perhaps this should mention that `info` key is on the `options` hash

lib/brainstem/dsl/base_block.rb
juanca: I'm thinking if the user provided two hashes: 1. for the description as `{ info: ... }` 2. for the options `{ ... }`. Since we're still supporting a third argument, it's possible users will rely to heavily on invalid args checks from ruby.

PR #98, opened 2017-08-24
Fix Filter and Search

JoeGaebel:
Filter And Search was clobbering the order of search results in two places.

Initially it would:
1. Run search and get `ordered_search_ids`
2. Instantiate a scope with `scope.where(id: ordered_search_ids)`
3. Apply filters
4. Apply at least a default ordering
5. Return the models to the frontend

2 and 4 would destroy the order of relevancy that ElasticSearch returned.

This PR determines whether or not the user wants to `order` their search results based on their params. In the case that the user does want order, it imposes that order.

Otherwise, it searches, filters the results, and then reorders them using the `ordered_search_ids`.

Related to https://github.com/mavenlink/mavenlink/pull/9184

### DIFF
--- a/lib/brainstem/query_strategies/base_strategy.rb
+++ b/lib/brainstem/query_strategies/base_strategy.rb
@@ -58,7 +58,7 @@ module Brainstem
         end
       end
 
-      def order_for_search(records, ordered_search_ids, with_models = true)
+      def order_for_search(records, ordered_search_ids, options = {})
         ids_to_position = {}
         ordered_records = []
 
@@ -67,7 +67,7 @@ module Brainstem
         end
 
         records.each do |record|
-          ordered_records[ids_to_position[with_models ? record.id : record ]] = record
+          ordered_records[ids_to_position[options[:with_ids] ? record : record.id]] = record
         end
 
         ordered_records.compact
--- a/lib/brainstem/query_strategies/base_strategy.rb
+++ b/lib/brainstem/query_strategies/base_strategy.rb
@@ -57,6 +57,33 @@ module Brainstem
           end
         end
       end
+
+      def order_for_search(records, ordered_search_ids)
+        ids_to_position = {}
+        ordered_records = []
+
+        ordered_search_ids.each_with_index do |id, index|
+          ids_to_position[id] = index
+        end
+
+        records.each do |record|
+          ordered_records[ids_to_position[record.id]] = record
+        end
+
+        ordered_records.compact
+      end
+
+      def calculate_limit_and_offset
+        if @options[:params][:limit].present? && @options[:params][:offset].present?
+          limit = calculate_limit
+          offset = calculate_offset
+        else
+          limit = calculate_per_page
+          offset = limit * (calculate_page - 1)
+        end
+
+        [limit, offset]
+      end
     end
   end
 end
--- a/lib/brainstem/query_strategies/base_strategy.rb
+++ b/lib/brainstem/query_strategies/base_strategy.rb
@@ -58,7 +58,7 @@ module Brainstem
         end
       end
 
-      def order_for_search(records, ordered_search_ids)
+      def order_for_search(records, ordered_search_ids, with_models = true)
         ids_to_position = {}
         ordered_records = []
 
@@ -67,7 +67,7 @@ module Brainstem
         end
 
         records.each do |record|
-          ordered_records[ids_to_position[record.id]] = record
+          ordered_records[ids_to_position[with_models ? record.id : record ]] = record
         end
 
         ordered_records.compact
--- a/lib/brainstem/query_strategies/filter_and_search.rb
+++ b/lib/brainstem/query_strategies/filter_and_search.rb
@@ -38,7 +38,8 @@ module Brainstem
 
         result_ids, _ = @options[:primary_presenter].run_search(@options[:params][:search], search_options)
         if result_ids
-          [scope.where(id: result_ids), result_ids]
+          resulting_scope = scope.where(id: result_ids)
+          [resulting_scope, result_ids]
         else
           raise(SearchUnavailableError, 'Search is currently unavailable')
         end
--- a/lib/brainstem/query_strategies/filter_and_search.rb
+++ b/lib/brainstem/query_strategies/filter_and_search.rb
@@ -39,14 +39,7 @@ module Brainstem
       end
 
       def paginate(scope)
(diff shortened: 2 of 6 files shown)


lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -28,23 +38,25 @@ def run_search(scope, includes)
...
         scope.limit(limit).offset(offset).distinct
       end
+
+      def ordering?
+        @options[:params][:order].present?
+      end
+
+      def paginate_array(array)
+        limit, offset = calculate_limit_and_offset
+        array.drop(offset).first(limit) # do we need to uniq this?
```
JoeGaebel: Still need to look into this

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -2,24 +2,34 @@ module Brainstem
...
         scope = @options[:primary_presenter].apply_filters_to_scope(scope, @options[:params], @options)
-        scope = @options[:primary_presenter].apply_ordering_to_scope(scope, @options[:params])
-        count = scope.count
-        scope = paginate(scope)
-        primary_models = evaluate_scope(scope)
+
+        if ordering?
+          count = scope.count
+          scope = paginate(scope)
+          scope = @options[:primary_presenter].apply_ordering_to_scope(scope, @options[:params])
```
bennapp: When we are ordering, shouldn't we apply ordering and then paginate?

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -28,23 +38,25 @@ def run_search(scope, includes)
 
         result_ids, _ = @options[:primary_presenter].run_search(@options[:params][:search], search_options)
         if result_ids
-          result_ids
+          [scope.where(id: result_ids), result_ids]
```
bennapp: Help me understand, what is this for?

JoeGaebel: This just instantiates the scope and returns the ordered array, makes it a little nicer than doing an extra assignment, and I can do `scope, ordered_ids = this_method_call(args_n_stuff)`

lib/brainstem/query_strategies/filter_and_search.rb
JoeGaebel: Oooo. Yes.

JoeGaebel: Yeah, but `paginate` in this case only adds AREL things to the SQL query and it doesn't actually cut out the pages we want in that line.

bennapp: Nice, sounds good.

lib/brainstem/query_strategies/filter_and_search.rb
bennapp: NON-BLOCKING:

Maybe two line this?

```
resulted_models = scope.where(id: result_ids)
[resulted_models, result_ids]
```

Idk if `resulted_models` is the best name but you get it.

JoeGaebel: Sure man!

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -28,23 +38,28 @@ def run_search(scope, includes)
...
-      def paginate(scope)
-        if @options[:params][:limit].present? && @options[:params][:offset].present?
-          limit = calculate_limit
-          offset = calculate_offset
-        else
-          limit = calculate_per_page
-          offset = limit * (calculate_page - 1)
-        end
+      def ordering?
+        sort_name = @options[:params][:order].to_s.split(":").first
```
urnf: Double check that you always have `[:order]` off `[:params]`

JoeGaebel:
```
Ruby:(2.3.3) object:(...FilterAndSearch>) >> @options[:params]
=> {"owned_by"=>"1", :search=>"toot, otto, toot"}
Ruby:(2.3.3) object:(...FilterAndSearch>) >> @options[:params][:order]
=> nil
Ruby:(2.3.3) object:(...FilterAndSearch>) >> @options[:params][:order].to_s
=> ""
Ruby:(2.3.3) object:(...FilterAndSearch>) >> @options[:params][:order].to_s.split(":")
=> []
Ruby:(2.3.3) object:(...FilterAndSearch>) >> @options[:params][:order].to_s.split(":").first
=> nil
```

I think we're good :+1:

urnf: Cool

lib/brainstem/query_strategies/base_strategy.rb
```diff
@@ -57,6 +57,33 @@ def filter_includes
...
+      def order_for_search(records, ordered_search_ids, with_models = true)
+        ids_to_position = {}
+        ordered_records = []
+
+        ordered_search_ids.each_with_index do |id, index|
+          ids_to_position[id] = index
+        end
+
+        records.each do |record|
+          ordered_records[ids_to_position[with_models ? record.id : record ]] = record
```
jhmoore: Extra space in the array here

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -2,24 +2,34 @@ module Brainstem
...
+          primary_models = evaluate_scope(scope)
+        else
+          filtered_ids = scope.pluck(:id)
+          count = filtered_ids.size
+
+          ordered_ids = order_for_search(filtered_ids, ordered_search_ids, false)
+          ordered_paginated_ids = paginate_array(ordered_ids)
+
+          scope = scope.unscoped.where(id: ordered_paginated_ids)
+          primary_models = scope.to_a
```
jhmoore: I recommend leaving a comment here as to why we're _not_ using `evaluate_scope` in this case

lib/brainstem/query_strategies/base_strategy.rb
JoeGaebel: What is this `Owner` nonsense

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -2,24 +2,34 @@ module Brainstem
...
+          primary_models = order_for_search(primary_models, ordered_paginated_ids)
```
jhmoore: Also recommend a comment as to why we have to order twice, once again after the `where`

PR #99, opened 2017-11-09
Add `plural` to Brainstem::DSL::Association

naomiajacobs:
Part of the effort in: https://github.com/mavenlink/mavenlink/pull/10271

This PR adds a key to the association object called `plural` that is set on initialization. If a `plural` option is passed in, it defaults to that, otherwise it tries to figure out whether the association is plural or not by seeing if it can be singularized. This option can be used to find out whether an association is plural, and allows for an override in case it's not easily guessable from the name of the association.

### DIFF
--- a/spec/brainstem/dsl/association_spec.rb
+++ b/spec/brainstem/dsl/association_spec.rb
@@ -24,6 +24,33 @@ describe Brainstem::DSL::Association do
     end
   end
 
+  describe 'plural' do
+    context 'when `plural` is specified in the options' do
+      let(:options) { { plural: true } }
+
+      it 'returns the value specified in the plural key' do
+        expect(association.plural).to eq true
+      end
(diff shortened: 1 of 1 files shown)


naomiajacobs:
@bennapp can you review please?

spec/brainstem/dsl/association_spec.rb
```diff
@@ -24,6 +24,33 @@
...
+      let(:options) { { plural: true } }
+
+      it 'returns the value specified in the plural key' do
+        expect(association.plural).to eq true
+      end
+    end
+
+    context 'when `plural` is not specified in the options' do
+      it 'delegates to ruby singularize and pluralize to guess if the association is plural or singular' do
+        class Dress
```
bennapp: 👗

PR #100, opened 2017-12-14
Add Pagination data to API responses

ahuth:
RFC -> https://github.com/mavenlink/rfc/pull/28
Bigmaven PR -> https://github.com/mavenlink/mavenlink/pull/10740
Brainstem-js PR -> https://github.com/mavenlink/brainstem-js/pull/121
Brainstem-redux PR -> https://github.com/mavenlink/brainstem-redux/pull/55

Adds pagination information to API responses. Currently deciding between doing this or calculating this client-side. Client-side calculations are doable, but require holding on to some state (what `perPage` and `page` the requests were made with).

### DIFF
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -76,11 +76,14 @@ module Brainstem
 
       struct = {
         'count' => count,
-        'page_number' => count > 0 ? options[:params].fetch(:page, 1).to_i : 0,
-        'page_count' => count > 0 ? (count.to_f / page_size).ceil : 0,
-        'page_size' => page_size,
         'results' => [],
         brainstem_key => {},
+        '_meta' => {
+          'count' => count,
+          'page_count' => count > 0 ? (count.to_f / page_size).ceil : 0,
+          'page_number' => count > 0 ? options[:params].fetch(:page, 1).to_i : 0,
+          'page_size' => page_size,
+        }
       }
 
       # Build top-level keys for all requested associations.
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -76,7 +76,7 @@ module Brainstem
 
       struct = {
         'count' => count,
-        'page_number' => count > 0 ? params.fetch(:page, 1) : 0,
+        'page_number' => count > 0 ? params.fetch(:page, 1).to_i : 0,
         'page_count' => count > 0 ? (count.to_f / per_page).ceil : 0,
         'results' => [],
         brainstem_key => {},
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -71,11 +71,13 @@ module Brainstem
       selected_associations = filter_includes(options)
 
       optional_fields = filter_optional_fields(options)
+      page_size = per_page(options[:params])
 
       struct = {
         'count' => count,
         'page_number' => page_number(count, options[:params]),
-        'page_count' => page_count(count, per_page(options[:params])),
+        'page_count' => page_count(count, page_size),
+        'page_size' => page_size,
         'results' => [],
         brainstem_key => {},
       }
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -71,12 +71,11 @@ module Brainstem
       selected_associations = filter_includes(options)
 
       optional_fields = filter_optional_fields(options)
-      per_page = [1, options[:params].fetch(:per_page, default_per_page).to_i].max
 
       struct = {
         'count' => count,
-        'page_number' => count > 0 ? options[:params].fetch(:page, 1).to_i : 0,
-        'page_count' => count > 0 ? (count.to_f / per_page).ceil : 0,
+        'page_number' => page_number(count, options[:params]),
+        'page_count' => page_count(count, per_page(options[:params])),
         'results' => [],
         brainstem_key => {},
       }
@@ -189,6 +188,18 @@ module Brainstem
       if options[:as].present?
         raise "PresenterCollection#presenting no longer accepts the :as option.  Use the brainstem_key annotation in your presenters instead."
       end
+    end
+
+    def page_number(count, params)
+      count > 0 ? params.fetch(:page, 1).to_i : 0
+    end
+
+    def page_count(count, per_page)
+      count > 0 ? (count.to_f / per_page).ceil : 0
+    end
+
+    def per_page(params)
+      [1, params.fetch(:per_page, default_per_page).to_i].max
     end
   end
 end
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -71,8 +71,16 @@ module Brainstem
       selected_associations = filter_includes(options)
 
       optional_fields = filter_optional_fields(options)
+      params = options.fetch(:params, {})
+      per_page = [1, params.fetch(:per_page, default_per_page).to_i].max
 
-      struct = { 'count' => count, brainstem_key => {}, 'results' => [] }
+      struct = {
+        'count' => count,
+        'page_number' => count > 0 ? params.fetch(:page, 1) : 0,
+        'page_count' => count > 0 ? (count.to_f / per_page).ceil : 0,
+        'results' => [],
+        brainstem_key => {},
+      }
 
       # Build top-level keys for all requested associations.
       selected_associations.each do |association|
--- a/lib/brainstem/presenter_collection.rb
+++ b/lib/brainstem/presenter_collection.rb
@@ -71,7 +71,7 @@ module Brainstem
       selected_associations = filter_includes(options)
 
       optional_fields = filter_optional_fields(options)
-      page_size = per_page(options[:params])
+      page_size = per_page(options)
 
       struct = {
         'count' => count,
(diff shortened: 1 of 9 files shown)


juanca:
`page_size`?

ahuth:
@juanca, even though that's in the RFC, I left it out of the response for now since we don't actually need it for anything (that I can think of).

lib/brainstem/presenter_collection.rb
```diff
@@ -71,8 +71,16 @@ def structure_response(presented_class, primary_models, count, options)
...
 
       optional_fields = filter_optional_fields(options)
-
-      struct = { 'count' => count, brainstem_key => {}, 'results' => [] }
+      params = options.fetch(:params, {})
+      per_page = [1, params.fetch(:per_page, default_per_page).to_i].max
+
+      struct = {
+        'count' => count,
+        'page_number' => count > 0 ? params.fetch(:page, 1).to_i : 0,
```
bennapp: Are there helpers for this somewhere?

bennapp:
I would lean towards adding in `page-size` seems like a nice to have.

ahuth:
@bennapp @shirish-pampoorickal @jhmoore @juanca, can you gents take a look at this PR, please?

Ben, I decided to leave out page-size for now, as I couldn't come up with a use case for it. Maybe a use case will become more clear as we move forward?

ahuth:
@bennapp , thanks for the feedback regarding `page_size`. That's now been added. You're right that it makes sense to include it when we consider the idea that HTTP request/responses should be stateless. 👍

lib/brainstem/presenter_collection.rb
```diff
@@ -183,5 +191,17 @@ def check_for_old_options!(options)
...
+    def page_number(count, params)
+      count > 0 ? params.fetch(:page, 1).to_i : 0
+    end
+
+    def page_count(count, per_page)
+      count > 0 ? (count.to_f / per_page).ceil : 0
+    end
+
+    def per_page(params)
+      [1, params.fetch(:per_page, default_per_page).to_i].max
```
ahuth: Hmm... wonder if we need to be considering the max per page here, as well

juanca: That's a good call. I would think so. The should be a getter for that. We would need to reference it above as well.

bennapp: Nice catch. If we haven't already, we certainly want to guard against this.

spec/brainstem/presenter_collection_spec.rb
```diff
@@ -159,6 +159,73 @@ class MyException < Exception; end
...
+              let(:max_per_page) { max_per_page }
+
+              before do
+                @presenter_collection.default_per_page = default_per_page
+                @presenter_collection.default_max_per_page = default_max_per_page
+              end
+
+              it "handles '#{message}'" do
+                expect(result["page_size"]).to eq(expected)
+              end
```
ahuth: @shirish-pampoorickal @bennapp @juanca , I'm not loving this test...

But I **_really_** don't want to write out all the different cases manually. Any thoughts or other ideas?

lib/brainstem/presenter_collection.rb
```diff
@@ -183,5 +191,11 @@ def check_for_old_options!(options)
         raise "PresenterCollection#presenting no longer accepts the :as option.  Use the brainstem_key annotation in your presenters instead."
       end
     end
+
+    def calculate_per_page(options)
+      per_page = [(options[:params][:per_page] || options[:per_page] || default_per_page).to_i, (options[:max_per_page] || default_max_per_page).to_i].min
+      per_page = default_per_page if per_page < 1
+      per_page
+    end
```
ahuth: This method is copied [_almost_ verbatim from the base strategy](https://github.com/mavenlink/brainstem/blob/69426421d26258451bec1d4cc743c15e6020995e/lib/brainstem/query_strategies/base_strategy.rb#L39).

I experimented with writing a `Pagination.calculate_per_page` thing, but it always ended up feeling really wrong.

Any thoughts or other ideas, @shirish-pampoorickal @bennapp @juanca ?

lib/brainstem/presenter_collection.rb
```diff
@@ -71,8 +71,16 @@ def structure_response(presented_class, primary_models, count, options)
       selected_associations = filter_includes(options)
 
       optional_fields = filter_optional_fields(options)
+      page_size = calculate_per_page(options)
 
-      struct = { 'count' => count, brainstem_key => {}, 'results' => [] }
+      struct = {
+        'count' => count,
+        'page_number' => count > 0 ? options[:params].fetch(:page, 1).to_i : 0,
```
ahuth: If there are no results, should the page_number be 0, 1, or does it matter?

juanca: I think `0` makes the most sense. It's the same dealio with `count`. There are 0 pages, there are 0 results.

lib/brainstem/presenter_collection.rb
juanca: It looks like all we're missing is passing in the defaults as part of the options? 

I think sharing the code will be highly important. What felt wrong about extracting into a `Pagination` module?

spec/brainstem/presenter_collection_spec.rb
juanca: I think this is fine. 

I guess a more important question(s) to ask: 

- Why do we need to test all these cases? 
- Why can we not guarantee a well behaved pagination object?

lib/brainstem/presenter_collection.rb
ahuth: In my work on sharing the code, I either had to:

- Accept a large parameter list (per_page, default_per_pax, max_per_page, default_max_per_page).
- Accept the same options hash that's used here, which has a nested `params` key that may have some stuff in it, or may not.
- Accept two params (per_page, max_per_page) and leave usages of the shared function to deal with the default value stuff.

All of these felt unfortunate

juanca: Why are Pagination defaults defined in multiple places? Should the Pagination module be responsible for taking (just) user input and sanitizing?

ahuth: Maybe I'll leave the calculations in the strategy and pass the strategy around this file 🤔

spec/brainstem/presenter_collection_spec.rb
ahuth: I was thinking we'd want to test all the logic of the calculation. Also, not sure what you mean be guaranteeing a well-behaved pagination object.

juanca: If the presenter was using a Pagination module/object and that *thing* always enforced sane pagination values then we wouldn't need to deal with all these test cases.

ahuth: Right, that makes sense. We would still want to test all the cases somewhere, though, just not in here.

I think we're on the same page, though

lib/brainstem/presenter_collection.rb
juanca: It's not clear to me why option 3 wasn't a good one:

> Accept two params (per_page, max_per_page) and leave usages of the shared function to deal with the default value stuff.

---

Perhaps you can elaborate on what you tried to do for each of options with code snippets? 

With more eyes, we can collaboratively reach a good abstraction or a bad one, but at least we'll have more context on it.

ahuth: Yep, makes sense @juanca. For number 3, that ends up removing most of the logic from the calculation.

Without the default and max and default max per page handling, the calculation would be:

```rb
[per_page, max_per_page].min
```

And at that point, it didn't feel necessary to have an object to do the calcuation.

Maybe this is a sign that the real problem here is deconstructing the `options`, and not performing the `per_page` calculation?

spec/brainstem/presenter_collection_spec.rb
juanca: I think we can drop the `nil` cases by using defaults and I think we can drop the invalid states by using validations. 

The tests can be more streamlined and get "to the point" by only asserting simple transformations (e.g. given a `per_page` bigger than the max will always set it to the max) versus these integration tests. 

That said: I may not have all the context on all these states and tests. All of these comments are more conversation starters. Would appreciate other contributors chiming in.

ahuth: We need the `nil` cases because there's a bunch of default handling going on.

For example, if no `per_page` is provided, then the `default_per_page` is used. And if no `max_per_page` is provided, there's a `default_max_per_page` that gets used.

The nils end up being important to the logic. Now, if the logic gets moved out of here they don't need to say in here, but they need to be tested somewhere.

ahuth:
@bennapp @shirish-pampoorickal can you gents take a look at this, please? Should be ready to go now, and Val and I will be QA-ing this thing heavily.

spec/shared/base_strategy.rb
```diff
@@ -0,0 +1,36 @@
+require "spec_helper"
+
+shared_examples_for Brainstem::QueryStrategies::BaseStrategy do
+  let(:strategy) { described_class.new(options) }
+
+  describe "#calculate_per_page" do
+    let(:result) { strategy.calculate_per_page }
+
+    [
+      # per_page  default_per_page  max_per_page  default_max_per_page  expected   message
```
bennapp: This is pretty cool

bennapp: A nice to have, and maybe something for next time would be dynamic `it` descriptions. Here we have dynamic describes, ie. `no per page and default > max` but then it would be nice to add something for the `it`, ie. `it uses max per page`.

ahuth: That's a good call, @bennapp. I'll look at adding that, as we have a day or two before deploying this

PR #102, opened 2017-12-27
Detect the count in MySQL

racerpeter:
Builds off of of https://coderwall.com/p/n3b0iw/mysql-sql_calc_found_rows-for-pagination-performance to introduce MySQL's [FOUND_ROWS()](https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows) functionality via a new Brainstem config called `Brainstem.mysql_use_calc_found_rows`.

This avoids an additional round trip to the MySQL server, and a second potentially expensive query to calculate the count.

**Best Case Scenario for Performance Improvement**
In a base scope where the paginated data query (with `LIMIT`/`OFFSET`) and the `COUNT` query take an equal amount of time to execute (for example, when the query does a full table scan), then the Brainstem response time would be cut in half.

**Worse Case Scenario for Performance Improvement**
In a base scope where either the paginated query or the count query were significantly faster than the other (for example, if the query does a full table scan, but an index can be used to determine record order, then the count would be significantly slower than a paginated query on the first few pages of results). In this instance, the improvement would be more modest-- eliminating a small amount of query time, and a round trip to the server.


racerpeter:
@cantino any thoughts on this one?

racerpeter:
@diclophis here you go-- can I get a re-review?

racerpeter:
Also, @jhmoore I hope you're happy.

... with the specs.

cantino:
Seems like a nice improvement @racerpeter!

PR #106, opened 2018-02-06
Add required & type options to endpoint params

shirish-pampoorickal:
The following features have been added:

- Endpoint parameters accepts `required` option. If a parameter is specified as required, the markdown generators will add `Required: true` to the generated documentation for that parameter
- Endpoint parameters accepts `type` and `item` options where type is the `type` of the parameter whereas `item` is the type of the sub item. eg. For an array of ids, you specify the type as `array` & item as `integer`.
- Add support for nested parameters on an endpoint
- Add support for markdown generation of nested parameters on an endpoint

```
  actions :create do
    valid :source, info: 'the source of the message'
    model_params :post do |post|
      post.valid :message, info: "the title of the post", required: true
      post.valid :reactions, type: 'array', item: 'string', info: 'an array of reactions'
      post.valid :permission, type: 'hash', info: 'permissions for the message' do |permission_param|
        permission_param.valid :editable, type: 'boolean', info: 'indicates if the message can be edited'
        permission_param.valid :viewable_by, type: 'array', item: 'integer'
      end
      post.valid :replies, type: 'array', item: 'hash' do |reply_param|
        reply_param.valid :message, info: "the message of the reply", required: true
        reply_param.valid :user_id, type: 'integer', info: "the id of the creator of the reply"
      end
    end
  end
```

Generates the following documentation:
```
##### Valid Parameters

- `source` (`String`) - which ids to include
- `post` (`Hash`)
    - `message` (`String`) - the message of the post
        - Required: true
    - `reactions` (`Array<String>`) - an array of reactions
    - `permission` (`Hash`) - permissions for the message
        - editable (`Boolean`) - indicates if the message can be edited
        - viewable_by (`Array<Integer>`)
    - `replies` (`Array<Hash>`)
        - `message` (`String`) - the message of the reply
            - Required: true
        - `user_id` (`Integer`) - the id of the creator of the reply
```

NOTES:
- Ideally we would keep the format consistent for specifying the type like in a presenter
`field :message, :string` but this would be a breaking change & since the type option is specific to generating API docs, I decided to keep it optional.
- The required, type & item optional parameter will help generate more robust OAS.

Benefits:

- This will helps us build


ahuth:
Nice work, @shirish-pampoorickal! I'm still looking through the changes, and in the meantime, what do you think about

- Bumping the minor version number to 1.2.0
- Adding a changelog entry

?

shirish-pampoorickal:
@ahuth Doing that right now :)

juanca:
Just for clarification: Is the main motivation behind this effort to produce more details auto-docs? 

> This will helps us build

Missing a word :P

CHANGELOG.md
```diff
@@ -1,5 +1,49 @@
 # Changelog
 
++ **1.1.2** - _02/06/2018_
```
ahuth: @shirish-pampoorickal, v 1.1.2 works for me if that's what you want here. 👍 

If we want to be semantic here, though, we "should" bump the minor version number when adding a feature.

Doesn't really matter here. Just thought I'd point it out. Perhaps when we have OpenAPI specification stuff, we'd want to bump v2?

lib/brainstem/api_docs/formatters/markdown/helper.rb
```diff
@@ -69,6 +69,15 @@ def md_li(text, indent_level = 0)
           def md_a(text, link)
             "[#{text}](#{link})"
           end
+
+
```
juanca: Comment: It's so weird that the spacing is a double newline between each method.

lib/brainstem/api_docs/formatters/markdown/endpoint_formatter.rb
```diff
@@ -83,59 +83,63 @@ def format_endpoint!
...
-            options = options.dup
+            options = options.deep_dup
             text    = md_inline_code(title)
 
+            text += md_inline_type(options.delete(:type), options.delete(:item)) if options.has_key?(:type)
             text += " - #{options.delete(:info)}" if options.has_key?(:info)
 
             if options.keys.any?
               text += "\n"
+              text += md_li("Required: #{options[:required].to_s}",   indent + 1) if options.has_key?(:required) && options[:required]
```
juanca: Opinionated: I would pull out `indent + 1` and name it accordingly. It seems that it will always be the next indent.

lib/brainstem/api_docs/formatters/markdown/endpoint_formatter.rb
```diff
@@ -83,59 +83,63 @@ def format_endpoint!
...
+          # @option [String] options :type The type of the field.
+          #   e.g. string, integer, boolean, array, hash
+          # @option [String] options :item The type of the items in the field.
+          #   Ideally used when the type of the field is an array or hash.
           # @param [Integer] indent how many levels the output should be
           #   indented from normal
           #
           def parameter_with_indent_level(title, options = {}, indent = 0)
-            options = options.dup
+            options = options.deep_dup
```
juanca: It doesn't seem like `deep_dup` is necessary for this method. Everything seems to only be modifying the root-level options hash (`delete`)

lib/brainstem/api_docs/formatters/markdown/endpoint_formatter.rb
juanca: Perhaps the indent parameter should already be incremented? It looks a bit inconsistent in `format_param_tree`

lib/brainstem/api_docs/endpoint.rb
```diff
@@ -120,32 +120,61 @@ def valid_params
...
-              if data.has_key?(:root)
-                key  = data[:root].respond_to?(:call) ? data[:root].call(controller.const) : data[:root]
-                (hsh[key] ||= []) << field_name
+      # Returns a hash of all params nested under the specified root or
+      # parent fields along with their type, item type & children.
+      #
+      # @return [Hash{Symbol => Hash}] root keys and their type info, item info & children
+      #   nested under them.
+      #
+      def params_configuration_tree
```
juanca: Following the logic for this method is pretty tedious. Is it possible to break it down further into more helper methods and lay down the general flow in a comment here -- does this make sense as an actionable revision?

lib/brainstem/api_docs/formatters/markdown/helper.rb
shirish-pampoorickal: Just keeping the pattern consistent with Rob's work. Don't want to spend time removing double newlines everywhere.

juanca: I think spending a little time on this file would be beneficial in the future. Keeping a maintained codebase will help future contributors understand desirable patterns. Although I understand that combing through the codebase is time consuming, we should tackle the problem piecewise so we can just "turn on" linting in the future without a hitch.

shirish-pampoorickal:
Closing this PR for since we are going to make this a two part release:
- Add support for type, item_type & required attributes for Endpoint parameters & Presenter fields (https://github.com/mavenlink/brainstem/pull/108/)
- Add support for specifying multi nested fields which have properties

PR #108, opened 2018-02-16
Add support for specifying type, item type and required options for Controller / Presenter fields

shirish-pampoorickal:
- Add the capability to indicate an endpoint param is required with the `required` key in the options hash.
  ```
        params.valid :message, :text,
                     required: true,
                     info: "the message of the post"
  ```
  - Add support for specifying the type of an endpoint param. For an endpoint param that has type `Array`,
    type of list items can be specified using `item_type` key in the options hash.
  ```
        params.valid :viewable_by, :array,
                     item_type: :integer,
                     info: "an array of user ids that can access the post"
  ```
  - Add support for specifying the data type of an item for a presenter field using `item_type` key in the
    options hash when the field is of type `Array`.
  ```
        field :aliases, :array,
              item_type: :string,
              info: "an array of user ids that can access the post"
  ```
  - Include the type and item type when generating markdown documentation for endpoint params.  
  - Add support for generating markdown documentation when the `required` option is specified on an endpoint param.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Changelog
 
++ **1.1.2** - _02/14/2018_
+  - Added support for indicating an endpoint param is required with the `required` key in the options hash.
+  ```
+        params.valid :message,
+          info: "the message of the post",
+          required: true
+  ```
+  - Added support for specifying the type of an endpoint param with the `type` key in the options hash. For endpoint params
+  that are an Array, type of list items can be specified using `item` key in the options hash. 
+  ```
+        params.valid :viewable_by,
+          type: "array",
+          item: "integer",
+          info: "an array of user ids that can access the post"
+  ```
+  - Added support for generating markdown documentation with the `required`, `type` and `item` keys in the options hash for an endpoint param.
+
 + **1.1.1** - _01/15/2017_
   - Add `Brainstem.mysql_use_calc_found_rows` boolean config option to utilize MySQL's [FOUND_ROWS()](https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows) functionality to avoid issuing a new query to calculate the record count, which has the potential to up to double the response time of the endpoint.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,21 +1,21 @@
 # Changelog
 
-+ **1.1.2** - _02/14/2018_
-  - Added support for indicating an endpoint param is required with the `required` key in the options hash.
++ **1.2.0** - _02/14/2018_
+  - Add the capability to indicate an endpoint param is required with the `required` key in the options hash.
   ```
-        params.valid :message,
-          info: "the message of the post",
-          required: true
+        params.valid :message, :text,
+          required: true,
+          info: "the message of the post"
   ```
-  - Added support for specifying the type of an endpoint param with the `type` key in the options hash. For endpoint params
-  that are an Array, type of list items can be specified using `item` key in the options hash. 
+  - Add support for specifying the type of an endpoint param. For an endpoint param that has type `Array`,
+  type of list items can be specified using `item_type` key in the options hash. 
   ```
-        params.valid :viewable_by,
-          type: "array",
-          item: "integer",
+        params.valid :viewable_by, :array,
+          item_type: :integer,
           info: "an array of user ids that can access the post"
   ```
-  - Added support for generating markdown documentation with the `required`, `type` and `item` keys in the options hash for an endpoint param.
+  - Include the type and item type when generating markdown documentation for endpoint params.  
+  - Add support for generating markdown documentation when the `required` option is specified on an endpoint param.
 
 + **1.1.1** - _01/15/2017_
   - Add `Brainstem.mysql_use_calc_found_rows` boolean config option to utilize MySQL's [FOUND_ROWS()](https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows) functionality to avoid issuing a new query to calculate the record count, which has the potential to up to double the response time of the endpoint.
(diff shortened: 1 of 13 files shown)


shirish-pampoorickal:
Hey @cantino! This was implemented for documentation purposes but can be used to enforce on the App side. e.g. `validate_params!`. The generated SDK's from the Open Api Specification will use the required field to enforce validations.

lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -233,6 +238,41 @@ def title(text, options = { nodoc: false })
...
+        end
+
+        DEFAULT_DATA_TYPE = 'string'
+        private_constant :DEFAULT_DATA_TYPE
+
+
+        def deprecated_type_warning
+          ActiveSupport::Deprecation.warn(
+            'Please specify the `type` of the parameter as the second argument. If not specified, '\
+              'it will default to `:string`. This default behavior will be deprecated in the next major '\
```
bennapp: 👍

lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -233,6 +238,41 @@ def title(text, options = { nodoc: false })
...
+          end
+
+          type.to_s
+        end
+
+        DEFAULT_DATA_TYPE = 'string'
+        private_constant :DEFAULT_DATA_TYPE
+
+
+        def deprecated_type_warning
```
bennapp: 👍

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changelog
 
++ **1.2.0** - _02/14/2018_
```
ahuth: 👍 Thanks for updating the changelog

ahuth: @shirish-pampoorickal, do we need to bump the version in `version.rb` as well?

lib/brainstem/api_docs/formatters/markdown/helper.rb
```diff
@@ -69,6 +69,15 @@ def md_li(text, indent_level = 0)
           def md_a(text, link)
             "[#{text}](#{link})"
           end
+
+
```
ahuth: 👮extra newline?

lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -233,6 +238,41 @@ def title(text, options = { nodoc: false })
...
+          ActiveSupport::Deprecation.warn(
+            'Please specify the `type` of the parameter as the second argument. If not specified, '\
+              'it will default to `:string`. This default behavior will be deprecated in the next major '\
+              'version and will need to be explicitly specified. e.g. `post.valid :message, :text, required: true`',
```
ahuth: Heredoc might be easier for a string like this. Typing out all those backslashes is a pain, man!

lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -233,6 +238,41 @@ def title(text, options = { nodoc: false })
           configuration[brainstem_params_context][:title] = \
             options.merge(info: text)
         end
+
+
```
ahuth: Intentional extra newline? Perhaps we should think about bringing in our Rubcop lint rules to this repo.

spec/brainstem/api_docs/formatters/markdown/endpoint_formatter_spec.rb
```diff
@@ -184,6 +190,36 @@ module Markdown
                   it "includes the legacy status if specified" do
                     expect(subject.output).to include "Legacy: false"
                   end
+
+                  context "when required option is specified" do
```
ahuth: Any assertion we can make to ensure that this is true. Usually I go on to folks about how easy it is to end up in a different state than we think we're in (seen it so many times), but I know that you know what you're doing.

Good be as easy as

```rb
context "when required is true" do
  before do
    expect(show_config.dig(:valid_params, :sprocket_id, :required)).to eq(true)
  end
```

spec/brainstem/api_docs/formatters/markdown/helper_spec.rb
```diff
@@ -93,6 +93,25 @@ module Markdown
               expect(subject.md_a("text", "link.md")).to eq "[text](link.md)"
             end
           end
+
+
```
ahuth: Extra newline!?!?!? 😲

brainstem.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency "rr"
   gem.add_development_dependency "rspec", "~> 3.5"
   gem.add_development_dependency "sqlite3"
-  gem.add_development_dependency "mysql2"
+  gem.add_development_dependency "mysql2", "0.4.10"
```
juanca: Why is the a strict requirement? I think we should be supporting up to `5.7.17-13` (since that's my workstation's version ;) )

lib/brainstem/version.rb
```diff
@@ -1,3 +1,3 @@
 module Brainstem
-  VERSION = "1.1.1"
+  VERSION = "1.2.0"
```
juanca: :+1:

lib/brainstem/api_docs/formatters/markdown/helper.rb
shirish-pampoorickal: @juanca @ahuth I am try to keep the formatting consistent. Once the documentation generation effort is complete, I will replace the double line breaks as a separate PR

brainstem.gemspec
shirish-pampoorickal: Some of our gems are not compatible with the latest `mySql2` gem version so had to lock it

lib/brainstem/api_docs/formatters/markdown/helper.rb
juanca: Whichever way works with me in the short term. Long term, I would like to have a discussion about adding some linting ruleset for the sake of consistency.

brainstem.gemspec
juanca: Oh, the gem! I connected the wrong dots. 

This is fine. Currently we're on `0.4.10` as well.

PR #109, opened 2018-03-01
Add support for evaluable field blocks & nested params

shirish-pampoorickal:
- Add Block Fields(`Brainstem::DSL::FieldBlock`) that can be evaluated with the presented model. Field blocks are evaluable only if :dynamic, :via or :lookup properties are specified.
    ```ruby
        fields :tags, :array,
               info: "The tags for the given category",
               dynamic: -> (widget) { widget.tags } do |tag|

          tag.field :name, :string,
                    info: "Name of the assigned tag"
        end
    ```
- Sub fields of these blocks are evaluated with the result of the parent block.
- Fix issue with nested fields being unable to have the same name as top level fields.
- Add support for nested parameters on an endpoint

    ```ruby
      actions :create do
        valid :source, :string,
              info: 'the source of the message'

        model_params :post do |post|
          post.valid :message, :string,
                     info: "the title of the post", required: true

          post.valid :permission, :hash,
                     info: 'permissions for the message' do |permission_param|

            permission_param.valid :editable, :boolean,
                                   info: 'indicates if the message can be edited'
            permission_param.valid :viewable_by, :array,
                                   item_type: 'integer'
          end

          post.valid :replies, :array, item_type: :hash do |reply_param|

            reply_param.valid :message, :string,
                              info: "the message of the reply", required: true
            reply_param.valid :user_id, :integer,
                              info: "the id of the creator of the reply"
          end
        end
      end
    ```
- Support generation of markdown documentation for the nested block fields.
- Add support for markdown generation of nested parameters on an endpoint.
- *NOT DOCUMENTED / BETA FEATURE*: Sub fields can also support evaluating with presented model if `use_parent_value` option is set to `false`. NOTE: Looking for better names. Suggestion: `use_presented_model`


shirish-pampoorickal:
@ahuth @bennapp Appreciate a look

lib/brainstem/dsl/block_field.rb
```diff
@@ -0,0 +1,69 @@
...
+
+module Brainstem
+  module DSL
+    class BlockField < Field
+      attr_reader :configuration
+
+      delegate :to_h, :keys, :has_key?, :each, to: :configuration
+
+      def initialize(name, type, options, parent_field = nil)
+        super(name, type, options)
```
ahuth: Woah, this subclass takes a different number of args than the superclass?! Weird man

lib/brainstem/dsl/block_field.rb
```diff
@@ -0,0 +1,69 @@
...
+          proc = options[:dynamic]
+          if proc.arity == 1
+            helper_instance.instance_exec(model, &proc)
+          else
+            helper_instance.instance_exec(&proc)
+          end
+        elsif options[:via]
+          model.send(options[:via])
+        else
+          raise "Block field #{name} can only be evaluated if :dynamic, :lookup, :via options are specified."
```
ahuth: Nice, I like the `raise` so that people get immediate feedback.

lib/brainstem/dsl/field.rb
```diff
@@ -38,7 +38,12 @@ def optional?
         options[:optional]
       end
 
+      # Please override in sub classes to compute value of field with the given arguments.
```
ahuth: Interesting. So if people don't override this, it's likely to be wrong?

spec/brainstem/api_docs/formatters/markdown/presenter_formatter_spec.rb
```diff
@@ -110,39 +109,39 @@ module Markdown
 
 
             describe "#format_fields!" do
-              let(:valid_fields) { {} }
+              let(:presenter_class) do
+                Class.new(Brainstem::Presenter) do
+                  presents Workspace
+                end
+              end
```
ahuth: Nice! Good job avoiding polluting tests by defining a constant here

README.md
```diff
@@ -81,6 +81,38 @@ module Api
               info: "the time of this Widget's last update"
         field :created_at, :datetime,
               info: "the time at which this Widget was created"
+
+        # Fields can be nested under non-executable parent fields where the nested fields
```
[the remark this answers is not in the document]
shirish-pampoorickal: Might need a better word to explain this but `non-executable` means that the parent block is not evaluated and it's nested fields are evaluated with the presented model rather than the value from the parent block

lib/brainstem/dsl/field.rb
shirish-pampoorickal: The base class in this case `DSL::Field` needs to evaluated with the default `evaluate_value_on`. It's not necessary to override it but in the case of `DSL::BlockField` we override the method to make it behave differently

lib/brainstem/api_docs/formatters/markdown/endpoint_formatter.rb
```diff
@@ -83,36 +83,34 @@ def format_endpoint!
...
-                      param_name,
-                      endpoint.valid_params[param_name],
-                      1
-                    )
-                  end
-
-                  buff << md_li(text)
-                end
-
+              endpoint.params_configuration_tree.inject("") do |buff, (param_name, param_config)|
```
brandonduff: If we want to always return the initial object for the next iteration, `each_with_object` might be a better choice than `inject`. It will allow us to get rid of explicitly returning `buff` on line 92.

shirish-pampoorickal:
YOLO! Merging the PR

PR #113, opened 2018-04-12
Validate / Ignore Unknown Params

shirish-pampoorickal:
- Add the capability to validate params based on documented parameters on endpoints. Invoking
    `brainstem_validate_params!` raises an error if unknown parameters are encountered.
  - Add the capability to ignore unknown params based on documented parameters on endpoints. Invoking
    `brainstem_ignore_unknown_params!` returns a copy of params while excluding unknown parameters.

### DIFF
--- a/lib/brainstem/params_validator.rb
+++ b/lib/brainstem/params_validator.rb
@@ -1,0 +1,88 @@
+require 'brainstem/unknown_params'
+
+module Brainstem
+  class ParamsValidator
+    attr_reader :unknown_params, :sanitized_params
+
+    def self.validate!(action_name, input_params, valid_params_config, options = {})
+      new(action_name, input_params, valid_params_config, options).validate!
+    end
+
+    def initialize(action_name, input_params, valid_params_config, options = {})
+      @valid_params_config = valid_params_config
+      @options             = options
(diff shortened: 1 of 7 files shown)


brandonduff:
This is awesome. What else needs to happen for this to go in?

lib/brainstem/params_validator.rb
```diff
@@ -0,0 +1,92 @@
...
+      begin
+        self.class.validate!(@action_name, value, valid_nested_params)
+      rescue Brainstem::ValidationError => e
+        @unknown_params << { parent_param_key => e.unknown_params }
+        @malformed_params << { parent_param_key => e.malformed_params }
+      end
+    end
+
+    def sanitize_input_params!(input_params)
+      malformed_params_error! unless input_params.is_a?(Hash) && input_params.present?
```
brandonduff: Do we `to_h` params at some point before this? Otherwise this won't work in Rails 5, since `ActionController::Parameters` no longer descend from `Hash`.

shirish-pampoorickal:
@brandonduff I was not very happy with the formatting of the error messages on nested attributes. I would be happy to talk through it if you like.

PR #115, opened 2018-05-03
Engine Docs

ahuth:
Allow configuration of **_where_** to get routes from. Normally this will be `Rails.application`. However, we should also be able to configure an engine such as `Api::Engine`.

### DIFF
--- a/lib/brainstem/api_docs.rb
+++ b/lib/brainstem/api_docs.rb
@@ -101,7 +101,7 @@ module Brainstem
     # Is a string because most relevant classes are not loaded until much
     # later.
     #
-    # @see Brainstem::ApiDocs::RailsIntrospector#base_presenter_class=
+    # @see Brainstem::ApiDocs::RailsIntrospector#base_controller_class=
     #
     config_accessor(:base_controller_class) do
       "::ApplicationController"
--- a/lib/brainstem/api_docs.rb
+++ b/lib/brainstem/api_docs.rb
@@ -107,6 +107,18 @@ module Brainstem
       "::ApplicationController"
     end
 
+    #
+    # Defines the application or engine that all routes will be fetched from.
+    #
+    # Is a proc because most relevant classes are not loaded until much
+    # later.
+    #
+    # @see Brainstem::ApiDocs::RailsIntrospector#base_application_proc=
+    #
+    config_accessor(:base_application_proc) do
+      Proc.new { Rails.application }
+    end
+
 
     #
     # If associations on a presenter have no description, i.e. no documentation,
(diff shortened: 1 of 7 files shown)


ahuth:
@shirish-pampoorickal can you take a look at this PR when you get a chance?

It adds support for a `base_application_proc` configuration, as we discussed yesterday. Can't tell if this actually generates docs or not, though. To generate an OpenAPI specification, do we need a specific branch of Brainstem?

lib/brainstem/version.rb
```diff
@@ -1,3 +1,3 @@
 module Brainstem
-  VERSION = "1.3.0"
+  VERSION = "1.4.0"
```
ahuth: Note the minor version bump

spec/brainstem/api_docs/introspectors/rails_introspector_spec.rb
```diff
@@ -160,7 +158,38 @@ module Introspectors
...
+              subject do
+                described_class.with_loaded_environment(default_args.merge({ base_application_proc: base_application_proc }))
+              end
+
+              it "recognizes the configured application or engine's routes" do
+                expect(subject.routes.map { |route| route[:path] }).to match_array([
+                  { spec: "/fake_route_1" }.to_s,
+                  { spec: "/fake_route_2" }.to_s
+                ])
+              end
```
ahuth: This test is a little gnarly, but it's the cleanest thing I came up with. Basically, we configure the introspector with a `base_application_proc`, and test that we get the right routes out of it. This is similar to some of the other tests in here.

lib/brainstem/api_docs/introspectors/rails_introspector.rb
```diff
@@ -175,6 +176,24 @@ def base_controller_class
...
 
 
+        #
+        # Returns a proc that will return the application or engine to get routes from.
+        #
+        # @return [Proc] Proc that returns the Rails application or an engine
+        #
+        def base_application_proc
+          @base_application_proc ||= Proc.new { Rails.application }
+        end
```
ahuth: Duplicating the default value is a little suspect... but is also how the other methods in here do it.

lib/brainstem/api_docs.rb
```diff
@@ -101,12 +101,24 @@ module ApiDocs
     # Is a string because most relevant classes are not loaded until much
     # later.
     #
-    # @see Brainstem::ApiDocs::RailsIntrospector#base_presenter_class=
+    # @see Brainstem::ApiDocs::RailsIntrospector#base_controller_class=
```
shirish-pampoorickal: 👍

ahuth:
Thanks!

PR #116, opened 2018-05-09
Bugfix engine support

shirish-pampoorickal:
Allow configuration of where to get routes from. Normally this will be Rails.application. However, we should also be able to configure an engine such as `Api::Engine`.

### DIFF
--- a/lib/brainstem/api_docs/introspectors/rails_introspector.rb
+++ b/lib/brainstem/api_docs/introspectors/rails_introspector.rb
@@ -48,8 +48,7 @@ module Brainstem
 
 
         #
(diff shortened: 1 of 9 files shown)


lib/brainstem/api_docs/introspectors/rails_introspector.rb
```diff
@@ -47,6 +47,17 @@ def controllers
         end
 
 
+        #
+        # Returns the alternate application class or defaults to Rails.application class
+        # class.
+        #
+        # @return [Array<Class>] an array of descendant classes
```
shirish-pampoorickal: Fix

PR #118, opened 2018-08-02
Enhancements to Open API Specification v2 generation

shirish-pampoorickal:
Any feedback on the DSL for documenting associations, dynamic response keys, dynamic endpoint  params & nested arrays would be greatly appreciated

PR addresses the following:

#### Bugfixes:
- [X] As an API user, I should not see parent fields as required due to one of its subfields being required.
- [X] As an API user, I should not see top-level query params like `only `, `order `, `page`, `per_page`, `include` & `optional` when the endpoint is not returning brainstem models
- [X] As an API user, I can the see the required fields for a create / update (POST / PUT) request

#### Enhancements:
- [X] As an API user, I should be able to see model associations in the JSON response

  Automatically generate documentation for associations with the following:
  - `response_key` will be used if given, otherwise, it will use the name of the association appended with `_id` or `_ids` for `has_many`. Using the example below, the association for `PetCategory` will generate documentation for `shooby_id`, while the `Owner` association will generate documentation for `owner_id`.

  - The response will include references to all of the associations, including any `polymorphic_classes` added to polymorphic associations. In the example below, the documentation for the response will contain references to `PetCategory`, `Vaccine`, `Owner`, `Dog`, and `Cat`.

    ```ruby
    class PetsPresenter < Brainstem::Presenter
      presents Pet

      associations do
        association :pet_category, PetCategory,
                    response_key: :shooby_id,
                    type: :belongs_to
        association :vaccines, Vaccine,
                    response_key: :vaccine_ids,
                    type: :has_many
        association :owner, Owner,
                    response_key: :owner_id,
                    type: :has_one
        association :pettable, :polymorphic,
                    response_key: :pettable_ref,
                    polymorphic_classes: [Dog, Cat]
      end
    end
    ```

- [X] As an API developer, I should be able to document multi-nested arrays in parameters/responses.

    A response like `[[{a: "string", b: 1}, "another string", 4]]` can be described as follows
    ```ruby
    response :array, nested_level: 2, item_type: :hash do
      nested.field :a, :string
      nested.field :b, :integer
    end
    ```

- [X] As an API developer, I should be able to document responses with dynamic keys

    ```ruby
    {
        :attributes => { 
            :"#{name}_field" => 'world', 
            :"#{name}_ids" => [1, 2, 3], 
            :name => name
        },
        :"#{model.class_name}_klass" => {
            :"#{model.class_name}_identifier" => model.identifier,
            :"#{model.association_name}_ids" => [1, 2, 3]
        }
    }
    ```
    
    A response with non-static key fields written above can be described as follows:

    ```ruby
    class ContactsController < ApiController
      brainstem_params do
        actions :show do
          response :hash do |response_param|
            response_param.fields :attributes, :hash do |attributes|
              attributes.dynamic_key_field :string
              attributes.field :some_ids, :array, dynamic: true
              attributes.field :name, :string, required: true
            end

            response_param.dynamic_key_field :hash do |dk|
              dk.dynamic_key_field :string
              dk.field :other_ids, :array, dynamic: true
            end
          end
        end
      end
    end
    ```

- [X] As an API developer, I should be able to document dynamic request parameters

    A param with non-static keys like `{ contact.id => 10, :"#{contact.friends_association_name}_ids" => [1, 2, 3]  }` can be described as follows:

    ```ruby
    class ContactsController < ApiController
      brainstem_params do
        actions :create do
          valid :info, :hash, required: true do |param|
            param.valid_dynamic_param :integer, required: true
          
            param.valid :some_ids, :array, dynamic: true
          end
        end
      end
    end
    ```

- [X] As an API developer, I should be able to document sort orders that don't accept a direction
parameter (`asc`/`desc`). When direction is passed as false, no sort order is generated for the docs.

    ```ruby
    presenter_class.sort_order :relevance,
        info: "can't sort by irrelevance",
        direction: false
    ```

- [X] As an API developer, I can mark endpoints, presenters, parameters, & fields as internal. I can specify reasons why any of the abover were marked as internal.
    ```ruby
    class ContactsController < ApiController
      internal! "Only used internally"

      brainstem_params do
        actions :index do
          response :hash do
            field :count, :integer,
                  info: "Total count of contacts",
                  internal: "Internal eyes only"
            fields :contacts, :array,
                   item_type: :hash,
                   nodoc: true,
                   info: "Array of contact details" do
              field :full_name, :string,
                    info: "Full name of the contact",
                    internal: true
            end
          end
        end
      end
    end
    ```

- [X] As an API developer, I can generate an Open API Specification which includes internal endpoints, presenters, parameters, & fields.

    Generate Open APi Specification for everything that is not marked as `nodoc`:

    `brainstem generate --include-internal --open-api-specification=2`


spec/brainstem/concerns/controller_dsl_spec.rb
```diff
@@ -489,6 +489,137 @@ module Concerns
...
+            expect(param_keys[3].call).to eq('data')
+            details_data_key = param_keys[3]
+            details_data_config = valid_params[details_data_key]
+            expect(details_data_config[:root]).to be_nil
+            expect(details_data_config[:ancestors]).to eq([info_key])
+
+            expect(param_keys[4].call).to eq(dynamic_keyword)
+            dynamic_data_nested_config = valid_params[param_keys[4]]
+            expect(dynamic_data_nested_config[:root]).to be_nil
+            expect(dynamic_data_nested_config[:ancestors]).to eq([info_key, details_data_key])
```
dordan: Any reason we don't check the :dynamic_key on the config for the dynamic_data_nested_config?

lib/brainstem/api_docs/formatters/open_api_specification/version_2/endpoint/param_definitions_formatter.rb
```diff
@@ -239,23 +239,26 @@ def format_body_params
...
                     {
                       type:        'array',
                       title:       param_name.to_s,
                       description: format_description(param_config[:info]),
                       items: {
-                        type: 'object',
-                        properties: format_param_branch(nested_properties(param_data))
-                      }
+                        type:       'object',
+                        required:   required_children(nested_properties),
```
juanca: How is this being treated as a boolean instead of an array of keys?

juanca:
FYI: Using double quotes as the commit message delimiter is not recommended. Nested backticks tend to be interpreted as subshells.

e.g. 
```
$ git commit -m "Test: `hello`"
-bash: hello: command not found 
[branch SHA] Test: 
```

While using single quotes works as expected:

```
$ git commit -m 'Test: `hello`'
[branch SHA] Test: `hello` 
```

https://github.com/mavenlink/brainstem/pull/118/commits/d225df2d54d190f46d104b54a953caf93da37960

lib/brainstem/api_docs/formatters/open_api_specification/version_2/endpoint/param_definitions_formatter.rb
```diff
@@ -309,3 +281,4 @@ def format_param_branch(branch)
 
 Brainstem::ApiDocs::FORMATTERS[:parameters][:oas_v2] =
   Brainstem::ApiDocs::Formatters::OpenApiSpecification::Version2::Endpoint::ParamDefinitionsFormatter.method(:call)
+
```
ahuth: Do a lot of these files have extra newlines at the end, or am I seeing the diff wrong?

lib/brainstem/api_docs/formatters/open_api_specification/version_2/field_definitions/presenter_field_formatter.rb
```diff
@@ -0,0 +1,147 @@
...
+                    'description' => description,
+                    'items' => formatted_data
+                  }
+                else
+                  {
+                    'type' => 'array',
+                    'description' => description,
+                    'items' => format_nested_array_parent(nested_level - 1, formatted_data)
+                  }
+                end.with_indifferent_access.reject { |_, v| v.blank? }
```
ahuth: Do we have active support available to us? If so, we can probably get rid of these `reject { |_, v| v.blank? }`s and replace them with `.compact`

lib/brainstem/api_docs/formatters/open_api_specification/version_2/field_definitions/response_field_formatter.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+      end
+    end
+  end
+end
+
+Brainstem::ApiDocs::FORMATTERS[:response_field][:oas_v2] =
+  Brainstem::ApiDocs::Formatters::OpenApiSpecification::Version2::FieldDefinitions::ResponseFieldFormatter.method(:call)
+
+
```
ahuth: 👮

lib/brainstem/api_docs/formatters/open_api_specification/version_2/field_definitions/presenter_field_formatter.rb
```diff
@@ -0,0 +1,147 @@
...
+require 'brainstem/api_docs/formatters/abstract_formatter'
+require 'brainstem/api_docs/formatters/open_api_specification/helper'
+
+module Brainstem
+  module ApiDocs
+    module Formatters
+      module OpenApiSpecification
+        module Version2
+          module FieldDefinitions
+            class PresenterFieldFormatter < AbstractFormatter
```
juanca: A little bit of this class shares procedural methods with `ResponseFieldFormatter`.

Should `ResponseFieldFormatter`, `EndpointParameterFormatter`, and `PresenterFieldFormatter` instead share these methods?

lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -256,6 +256,31 @@ def fields(name, type, options = {}, &block)
           with_options(format_ancestry_options(formatted_name, field_block_config), &block)
         end
 
+        #
+        # Allows defining a dynamic key field block for a custom response
+        #
+        # @param [Symbol] type the data type of the response.
```
juanca: `:type` typo

spec/brainstem/concerns/controller_dsl_spec.rb
```diff
@@ -691,16 +691,22 @@ module Concerns
...
-                      dk.field :mk_blah, :string, required: true
+                      dk.dynamic_key_field :string,
+                        info: "I am a dynamic key field I"
+                      dk.field :mk_blah, :string,
+                        required: true
                     end
 
                     response_param.fields :mk2, :hash do |dk|
-                      dk.dynamic_key_field :string, required: true
+                      dk.field :_dynamic_key, :string,
```
juanca: We might want to use the constant `DYNAMIC_KEY`?

lib/brainstem/api_docs/formatters/open_api_specification/version_2/presenter_formatter.rb
```diff
@@ -52,18 +52,45 @@ def format_type!
...
             end
-            
+
+            def format_field_associations(properties)
+              presenter.valid_associations.each_with_object(properties) do |(_name, association), props|
+                if association.foreign_key
+                  case association.type
+                  when :belongs_to, :has_one
+                    props[association.foreign_key] = type_and_format(:integer).merge(description: association.description) unless props[association.foreign_key]
+                  when :has_many
```
juanca: > Rails supports six types of associations:
> - belongs_to
> - has_one
> - has_many
> - has_many :through
> - has_one :through
> - has_and_belongs_to_man

We might be missing three of these.

lib/brainstem/api_docs/formatters/open_api_specification/version_2/presenter_formatter.rb
```diff
@@ -52,18 +52,45 @@ def format_type!
...
+              branch.each_with_object(ActiveSupport::HashWithIndifferentAccess.new) do |(name, field), buffer|
                 buffer[name.to_s] = format_field(field)
-                buffer
               end
             end
-            
+
+            def format_field_associations(properties)
+              presenter.valid_associations.each_with_object(properties) do |(_name, association), props|
+                if association.foreign_key
```
juanca: When does an association not have `foreign_key` set?

juanca: Oh, I see this is part of the DSL -- instead of Rails Association objects.

lib/brainstem/api_docs/formatters/open_api_specification/version_2/presenter_formatter.rb
juanca: I see that we're only intentionally support three. Is this for the time being?

lib/brainstem/api_docs/controller.rb
```diff
@@ -101,13 +103,19 @@ def tag_groups
...
       end
 
       def valid_sorted_endpoints
         endpoints.sorted_with_actions_in_controller(const)
       end
+
+      private
+
+      def nodoc_for?(config)
+        !!(config[:nodoc] || (config[:internal] && !internal))
```
juanca: This is great! Now I can use our internal docs as a source of truth instead of just opting for source code. 👍

juanca:
I think we should probably DRY up the `nodoc` and `internal` keys.

It's really more like a 3-state key:

```
doc: public | private | internal
default: public
```

Isn't it confusing to have a field with both `nodoc` and `internal` keys?

juanca:
Another thing: I don't really understand the point behind the _dynamic_ changes. The examples don't really showcase the functionality. Can we elaborate on why it is needed?

spec/brainstem/api_docs/formatters/open_api_specification/version_2/endpoint/param_definitions_formatter_spec.rb
```diff
@@ -34,38 +34,86 @@ module Endpoint
...
+                        subject.call
+                      end
+                    end
 
-                  context 'when action is index' do
-                    let(:action) { 'index' }
+                    context 'when action is show' do
+                      let(:action) { 'show' }
+
+                      it 'formats path, optional, query and body params for the endpoint' do
```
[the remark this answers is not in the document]
shirish-pampoorickal: Commit Message: `Only add query and body params for endpoints that do not use a presenter to return data`

The commit message is correct. The intention was to say only query & body params are added to endpoints that do not use a brainstem presenter. It would have been better to also call out that `pagination`, `includes` & `only` (shared params) are not included on endpoints that do not use a brainstem presenter.

lib/brainstem/api_docs/formatters/open_api_specification/version_2/endpoint/param_definitions_formatter.rb
shirish-pampoorickal: This was a mistake. I will fix it.

lib/brainstem/api_docs/formatters/open_api_specification/version_2/field_definitions/presenter_field_formatter.rb
shirish-pampoorickal: Unfortunately, `compact` does not remove blank values :(

lib/brainstem/api_docs/formatters/open_api_specification/version_2/field_definitions/response_field_formatter.rb
shirish-pampoorickal: Fixed

spec/brainstem/concerns/controller_dsl_spec.rb
shirish-pampoorickal: We wanted to give a clear example of how the name of the field can be used to define a dynamic field

lib/brainstem/api_docs/formatters/open_api_specification/version_2/presenter_formatter.rb
shirish-pampoorickal: The type specified on the association is meant to determine the key name for the association.

In terms of having dynamic keys on the presented object, these six associations fall in two buckets. For `belongs_to`, `has_one ` & `has_one :through` will generate attribute name of `blah_id` whereas all the other ones will generate `blah_ids`

shirish-pampoorickal:
@juanca I have added some more text to explain the use of dynamic keys.

juanca:
I'll be honest: I kinda lost most of the context between these reviews. 

I'm a little bummed no one else has taken a look into these changes either.

Either way, I looked at the change set between my last comment and now and I don't see anything wrong. LGTM

shirish-pampoorickal:
> I'll be honest: I kinda lost most of the context between these reviews.
> 
> I'm a little bummed no one else has taken a look into these changes either.
> 
> Either way, I looked at the change set between my last comment and now and I don't see anything wrong. LGTM

@juanca I tried getting this reviewed from others but unfortunately, everyone is super busy :(

ahuth:
🎉 🍾 Nice work, Shirish! This one was a huge amount of effort over a long period of time! Not many people other than you could've pulled it off 😄

juanca:
Yeah! Good stuff! Definitely worth a champagne bottle.

PR #123, opened 2019-06-06
Prototype: process an additional param that is sent to associations

urnf:
# DO NOT MERGE.  MEANT FOR DISCUSSION.

Proof of concept meant to explore passing additional options to `dynamic` associations.  While this is tailored towards the placeholder "firstFiftyAssociations ", meant to address a current problem with pagination, something like this would lay the groundwork for additional options.

In the presenter, association definition goes from
```ruby
dynamic: lambda {|model| model.scope }
```
to
```ruby
dynamic: lambda {|model, options = {}| some_option_based_behavior ? model.something_else : model.scope }
```

The new param is not required to maintain backwards compatibility and leaves the responsibility of handling that option up to the association proc itself.  The procs themselves would probably end up needing to be refactored wholesale to support any new features that would be added to them, such as pagination.

### DIFF
--- a/lib/brainstem/dsl/association.rb
+++ b/lib/brainstem/dsl/association.rb
@@ -37,12 +37,14 @@ module Brainstem
         end
       end
 
-      def run_on(model, context, helper_instance = Object.new)
+      def run_on(model, context, helper_instance = Object.new, lookup_options = {})
         if options[:lookup]
           run_on_with_lookup(model, context, helper_instance)
         elsif options[:dynamic]
           proc = options[:dynamic]
-          if proc.arity == 1
+          if proc.arity == -2
+            helper_instance.instance_exec(model, lookup_options, &proc)
+          elsif proc.arity == 1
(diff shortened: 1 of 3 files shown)


lib/brainstem/dsl/association.rb
```diff
@@ -37,12 +37,14 @@ def method_name
...
       end
 
-      def run_on(model, context, helper_instance = Object.new)
+      def run_on(model, context, helper_instance = Object.new, lookup_options = {})
         if options[:lookup]
           run_on_with_lookup(model, context, helper_instance)
         elsif options[:dynamic]
           proc = options[:dynamic]
-          if proc.arity == 1
+          if proc.arity == -2
```
urnf: Since the expected association proc format takes in `model` and `options = {}`, we start with arity 2, and with one of the params not required, makes it -2.

I'm not sure about the case at the bottom with no arity - as far as I know, our documentation requires that at least a model is passed in, resulting in arity 1?

PR #124, opened 2019-06-13
Permit query params for presenting

brandonduff:


brandonduff:
Failures are unrelated to this change and seem to be hitting master. Fix here https://github.com/mavenlink/brainstem/pull/125

lib/brainstem/controller_methods.rb
```diff
@@ -36,7 +36,7 @@ def brainstem_present(name, options = {}, &block)
     #                           only required if the name cannot be inferred.
     # @return (see PresenterCollection#presenting)
     def brainstem_present_object(objects, options = {})
-      options.merge!(:params => params, :apply_default_filters => false)
+      options.merge!(:params => params.to_unsafe_h, :apply_default_filters => false)
```
naganowl: These concern me, but I don't have a better idea.

brandonduff: Because they’re only used for filtering, and not assigned to a model, I think it’s ok.

PR #129, opened 2019-10-14
Fix Empty has_many Association Presentation

kcotugno:
This fixes a bug that occurs when a dynamic/lookup association in a presenter returns nil. It is always treated as a has_one relationship. Thus the foreign key is incorrect, e.g., `workspace_id` instead of `workspace_ids`. Further the foreign key value is inconsistent with empty `Array`s and `ActiveRecord::Relation`s. So we now ensure that we return
an empty array if the value is nil.

### DIFF
--- a/lib/brainstem/presenter.rb
+++ b/lib/brainstem/presenter.rb
@@ -369,8 +369,8 @@ module Brainstem
           struct["#{singular_external_name}_ref"] = make_model_ref(associated_model_or_models)
         end
       else
-        if associated_model_or_models.is_a?(Array) || associated_model_or_models.is_a?(ActiveRecord::Relation)
-          struct["#{singular_external_name}_ids"] = associated_model_or_models.map { |associated_model| to_s_except_nil(associated_model.try(:id)) }
+        if associated_model_or_models.is_a?(Array) || associated_model_or_models.is_a?(ActiveRecord::Relation) || association.type == :has_many
+          struct["#{singular_external_name}_ids"] = associated_model_or_models&.map { |associated_model| to_s_except_nil(associated_model.try(:id)) } || []
         else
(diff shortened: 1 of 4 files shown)


lib/brainstem/presenter.rb
```diff
@@ -369,8 +369,8 @@ def add_ids_or_refs_to_struct!(struct, association, external_name, associated_mo
           struct["#{singular_external_name}_ref"] = make_model_ref(associated_model_or_models)
         end
       else
-        if associated_model_or_models.is_a?(Array) || associated_model_or_models.is_a?(ActiveRecord::Relation)
-          struct["#{singular_external_name}_ids"] = associated_model_or_models.map { |associated_model| to_s_except_nil(associated_model.try(:id)) }
+        if associated_model_or_models.is_a?(Array) || associated_model_or_models.is_a?(ActiveRecord::Relation) || association.type == :has_many
```
shirish-pampoorickal: Do we have this type available without someone explicitly specifying it as part of the documentation?

shirish-pampoorickal: From the spec, it looks like we are relying on documentation. Just an FYI, Documentation support is an optional feature supported by brainstem but ideally, we should find a solution that does not rely on documentation to fix a bug.

IMO, the bigmaven code is incorrect. It should not be responding with nil if are using a lookup for a has many association. Thoughts?

kcotugno: To answer your first question, no. The only way I can think of getting it is check if the name is plural. But that doesn't seem like a good way to check.

I understand that it is not required. I didn't realize that it was only for documentation. It does seem like good information to know if an association is a one or many.

I agree that the bigmaven code should be different. However, this seems to allow for a bug that is easy to introduce and hard to find.

PR #139, opened 2020-05-22
Row calc experiment

brentgaither:
Extract a method for checking if row calc is enabled.


lib/brainstem/query_strategies/base_strategy.rb
```diff
@@ -53,6 +53,12 @@ def get_ids_sql(scope)
         scope.klass.where(id: id_lookup.keys).sort_by { |model| id_lookup[model.id] }
       end
 
+      def use_calc_row?
+        return false unless Brainstem.mysql_use_calc_found_rows
+        return false unless ActiveRecord::Base.connection.instance_values["config"][:adapter] =~ /mysql/i
+        true
+      end
```
jarednielson, line 60: Is this just a mechanical refactor?

brentgaither: yes, it just gives me the ability to monkey patch it in bigmaven.

jhmoore:
🤘

PR #141, opened 2020-10-05
Add :if option to `valid`

jarednielson:
- When the if option is passed the method or proc is evaluated in the context of the controller instance
    - If the block returns true the param is considered valid and included in the valid_params_tree
    - If the block returns falsey the param is considered invalid and excluded from the valid_params_tree
  - This will allows users to dynamically determine if a brainstem param should be considered valid

### DIFF
--- a/lib/brainstem/concerns/controller_dsl.rb
+++ b/lib/brainstem/concerns/controller_dsl.rb
@@ -566,8 +566,7 @@ module Brainstem
           .to_h
           .inject(ActiveSupport::HashWithIndifferentAccess.new) do |hsh, (field_name_proc, field_config)|
 
-          if field_config.has_key?(:if)
-            condition = field_config[:if]
+          if (condition = field_config[:if])
             case condition
             when Symbol
               next hsh unless self.send condition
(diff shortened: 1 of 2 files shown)


lib/brainstem/concerns/controller_dsl.rb
```diff
@@ -564,6 +566,16 @@ def valid_params_tree(requested_context = action_name.to_sym)
           .to_h
           .inject(ActiveSupport::HashWithIndifferentAccess.new) do |hsh, (field_name_proc, field_config)|
 
+          if field_config.has_key?(:if)
```
brandonduff, line 569: Could use the [Safe Assignment in Condition idiom](https://github.com/rubocop-hq/ruby-style-guide#safe-assignment-in-condition) to reduce a line:

```
if (condition = field_config[:if])
  case condition
  ...
end
```

jarednielson: Nice call. I was just thinking the same thing.

bennapp:
Could you add examples / documentation to the README.md?

jarednielson:
Gonna close this because of potential issues w/ `brainstem-js` and block/allowed parameters.

PR #149, opened 2021-05-25
Allow enums to be reflected in OAS generation

cbacon91:
Enums are a legal part of the OAS data format. Brainstem does not allow a way for controllers / parameters to provide these enums though, so this PR is used to expose them.

https://swagger.io/docs/specification/data-models/enums/

By way of example, if the `access_level` in Workspaces:particpations is used as an enum, the only code change necessary is adding the `enum` key with the legal values in an array. The `type` of the field is still respected.

```ruby
participant.valid :access, :string,
                  enum: %w(collaboration time_logging financial admin),
  info: <<~INFO
    The access level of the participant. Must be one of the following: `collaboration`,
    `time_logging`, `financial` or `admin`.
  INFO
```

This generates the api documentation:
image

And finally the OAS Schema yml file:
```yml
participations:
  type: array
  description: Determines the participants in the project.
  items:
    type: object
    properties:
      user_id:
        type: integer
        format: int32
        description: The user ID of the participant.
      access:
        type: string
        description: |-
          The access level of the participant. Must be one of the following: `collaboration`,
          `time_logging`, `financial` or `admin`.
        enum:
          - collaboration
          - time_logging
          - financial
          - admin
```

### DIFF
--- a/lib/brainstem/api_docs/formatters/open_api_specification/version_2/field_definitions/response_field_formatter.rb
+++ b/lib/brainstem/api_docs/formatters/open_api_specification/version_2/field_definitions/response_field_formatter.rb
@@ -20,7 +20,7 @@ module Brainstem
               def format
                 field_config = @param_tree[:_config]
                 field_properties = @param_tree.except(:_config)
-  
+
                 format_field(field_config, field_properties)
               end
               alias_method :call, :format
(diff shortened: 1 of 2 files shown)


cpalafox (approved):
This is awesome! LGTM

shirish-pampoorickal (approved):
LGTM!

PR #150, opened 2021-08-02
Allow presenters to override count and id calculation from scope

jarednielson:
In BigMaven we need to cache the count value for `line_items` queries because getting the count is 90% of the compute time for each request. Since the use case is to paginate over all requests within a relatively short amount of time we can get the count and cache it in bigmaven. But we need brainstem to skip calculating the count.


shirish-pampoorickal (approved):
LGTM! Suggest bumping brainstem version

jarednielson:
> LGTM! Suggest bumping brainstem version

Done

PR #151, opened 2021-08-03
instance_exec count_evaluator within context of presenter helper

jarednielson:
Count evaluator block needs to be executed with the context of the presenter helper if one is present.


brentsondgeroth (approved):
LGTM

PR #152, opened 2021-08-09
Add seek pagination capability for better performance

dordan:
Leveraging seek pagination instead of a large offset and limit provides significantly better performance: https://www.eversql.com/faster-pagination-in-mysql-why-order-by-with-limit-and-offset-is-slow/


dordan:
We will need to implement dynamic scoping as the sort order might not always be ID.

PR #153, opened 2021-08-13
Use custom paginator when it is provided

rickdav:
What
---
* Use a provided paginator instead of the strategy's `paginate` method, falling back to original behavior if paginator 
  returns no results.
* Add warning when specs are not run for a given database.
* Unify the return object of paginate methods in strategies.

Why
---
A custom pagination strategy can improve performance on many endpoints, and this allows one to be used.

### DIFF
--- a/lib/brainstem/query_strategies/base_strategy.rb
+++ b/lib/brainstem/query_strategies/base_strategy.rb
@@ -6,32 +6,30 @@ module Brainstem
     class BaseStrategy
       def initialize(options)
         @options = options
-        @last_count = nil
       end
 
       def execute(scope)
         raise NotImplemented, 'Your strategy class must implement an `execute` method'
       end
 
-      def evaluate_scope(scope)
-        @last_count = nil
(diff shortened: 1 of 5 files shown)


lib/brainstem/query_strategies/base_strategy.rb
```diff
@@ -42,14 +40,35 @@ def calculate_per_page
 
       private
 
-      def get_ids_sql(scope)
+      def ids_and_count_for_page(page, scope, count_scope, should_paginate)
+        if use_calc_row? && @options[:paginator]
+          scope = primary_presenter.apply_ordering_to_scope(scope, @options[:params])
+          ids, count = @options[:paginator]&.get_paged(page, scope)
```
shirish-pampoorickal, line 46: `@options[:paginator]&` The safe operator might not be needed since you verified `@options[:paginator]` exists in the conditional.

rickdav:
This PR was re-implemented in https://github.com/mavenlink/brainstem/pull/159, which does a better job of separating responsibility.

PR #159, opened 2021-12-16
Use custom pagination strategy when it is provided

rickdav:
Use custom pagination strategy when it is provided

This is a re-implementation of an earlier PR: https://github.com/mavenlink/brainstem/pull/153

* Extract pagination logic into a separate class
* If a paginator is provided in options, use it instead of the default paginator.
* Add warning when specs are not run for a given database.


lib/brainstem/query_strategies/base_strategy.rb
```diff
@@ -64,7 +73,7 @@ def use_calc_row?
       end
 
       def delegate_count_to_presenter?
-        primary_presenter.evaluate_count?
+        primary_presenter&.evaluate_count?
```
ncperry, line 76: Can we avoid returning the nullable boolean? A method that has a question mark should return a true or false. It should not return a nil.

ncperry: On further review, what did we change that made the `primary_presenter` nullable? It doesn't seem like we made any change that would require that.

lib/brainstem/query_strategies/base_strategy.rb
```diff
@@ -42,17 +42,26 @@ def calculate_per_page
...
+        id_lookup = {}
+        ids.each.with_index { |id, index| id_lookup[id] = index }
+        scope.klass.where(id: id_lookup.keys).sort_by { |model| id_lookup[model.id] }
+      end
+
+      def get_ids_for_page(scope)
         if use_calc_row?
+          # The paginator uses mysql SQL_CALC_FOUND_ROWS.
+          ids = @options[:paginator]&.get_ids_for_page(calculate_page, scope)
+          return ids if ids.present?
```
ncperry, line 56: Could we get away with not nesting `@options[:paginator]` inside `use_calc_row?`? We may be creating a situation where people provide a paginator, and it doesn't get used, because `use_calc_row?` is false for some reason.

lib/brainstem/query_strategies/base_strategy.rb
```diff
@@ -29,7 +29,7 @@ def evaluate_scope(scope)
       def evaluate_count(count_scope)
         return primary_presenter.evaluate_count(count_scope) if delegate_count_to_presenter?
 
-        ret = @last_count || count_scope.count
+        ret = @last_count || @options[:paginator]&.get_count(count_scope) || count_scope.reorder(nil).count
```
ncperry, line 32: No one: Say you don't like null-coalescing without saying you don't like null-coalescing.
Nate:
```suggestion
        ret = begin
          return @last_count unless @last_count.nil?
          return paginator_override(count_scope) unless paginator_override.nil?
          count_scope.reorder(nil).count
        end
        @last_count = nil
        ret
      end
      
      def paginator_override
        @options[:paginator]
      end
```

ncperry: Also why did we have to add the `reorder(nil)` to the `count_scope`?

lib/brainstem/query_strategies/base_strategy.rb
rickdav: I went back and forth on this, and I agree with you. I went with the more cagey approach, but I'm happy to pull it out of the check, since the behavior of the paginator should be completely hidden from brainstem. It should be up to the creator of the paginator to make sure their runtime can handle their strategy.

lib/brainstem/query_strategies/base_strategy.rb
rickdav: Good catch. This was a quick add to fix an existing failing spec, and I didn't pay much attention to it.

lib/brainstem/query_strategies/base_strategy.rb
rickdav: I feel like the coalescing pattern is very idiomatic ruby - is there some downside in using it that I am unaware of?

Also this suggestion is a change of behavior such that `@last_count` will no longer always be set to nil. The `return` statements will return from the method, not the begin block.

brandonduff: I don't have a ton of context here so take this with a grain of salt, but if we have the idea of a `Paginator` that someone can provide, should we just have a default `Paginator` that Brainstem defines that has the same interface? (It's implementation would just be the `@last_count || count_scope.count` stuff). I think that may reveal intent and clean up the code a bit.

rickdav: @brandonduff - That's the thing I really wanted to reach for when starting this effort. I hit a lot of roadblocks along the way, and never revisited it once I understood the problem a bit better. I'll see if I can make something like that work.

rickdav: @ncperry - `reorder(nil)` will make sure that any ordering applied to the scope is removed, since ordering is not needed for a count query.

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -7,10 +7,9 @@ def execute(scope)
 
         if ordering?
           count_scope = scope
-          scope = paginate(scope)
+          scope = paginated_scope(scope)
           scope = @options[:primary_presenter].apply_ordering_to_scope(scope, @options[:params])
-          primary_models = evaluate_scope(scope)
-          count = evaluate_count(count_scope)
+          primary_models, count = paginator.paginate(page: calculate_page, scope: scope, count_scope: count_scope)
```
ncperry, line 12: So we `paginate` twice? It seems like the limit and offset that we're doing in FilterAndSearch should be a responsibility of the Paginator.

lib/brainstem/query_strategies/filter_or_search.rb
```diff
@@ -67,7 +63,7 @@ def run_search(scope, includes, sort_name, direction)
         end
       end
 
-      def paginate(scope)
+      def paginated_scopes(scope)
```
ncperry, line 66: Similarly it seems like this should be the responsibility of the Paginator not the Filterer or the Searcher.

lib/brainstem/query_strategies/paginator.rb
```diff
@@ -0,0 +1,69 @@
+module Brainstem
```
ncperry, line 1: I like your addition of a class here a lot.

spec/brainstem/query_strategies/paginator_spec.rb
```diff
@@ -0,0 +1,57 @@
+require "spec_helper"
+
+describe Brainstem::QueryStrategies::Paginator do
+  using_mysql = ActiveRecord::Base.connection.instance_values["config"][:adapter] =~ /mysql/i
+  if using_mysql
+    xit '- Tests requiring sqlite have not been run.' do
```
ncperry, line 6: Do we want the `xit`? Isn't that what the if-else is for?

spec/brainstem/query_strategies/paginator_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      expect(true).to eq false
```
ncperry, line 7: Could we possibly raise here instead of a true == false assertion? If we raise an error and give a good error message perhaps we can help anyone who runs into a failure here.

ncperry (approved):
LGTM! There is perhaps an SRP violation on what belongs in FilterAndSearch vs Paginator. But I'm comfortable merging this in to see if it works, then going back and spending more time on it!

spec/brainstem/query_strategies/paginator_spec.rb
rickdav: Yeah, I like that a lot!

spec/brainstem/query_strategies/paginator_spec.rb
rickdav: This is to give feedback to whoever is running the tests that there is a whole branch of tests that have been missed. I uncovered code that had failing tests for some time but that branch of tests had not been run before being merged.

lib/brainstem/query_strategies/filter_and_search.rb
```diff
@@ -3,14 +3,15 @@ module QueryStrategies
...
           count_scope = scope
-          scope = paginate(scope)
-          scope = @options[:primary_presenter].apply_ordering_to_scope(scope, @options[:params])
-          primary_models = evaluate_scope(scope)
-          count = evaluate_count(count_scope)
+          scope = primary_presenter.apply_ordering_to_scope(scope, @options[:params])
+          limit, offset = calculate_limit_and_offset
+          ids = paginator.get_ids(limit: limit, offset: offset, scope: scope)
+          count = paginator.get_count(count_scope)
+          primary_models = Brainstem::QueryStrategies::DataMapper.new.get_models(ids: ids, scope: scope)
```
ncperry, line 14: We could extract a `data_mapper` function here and in `filter_or_search`. We could put it in the `base_strategy` class.

spec/brainstem/query_strategies/filter_and_search_spec.rb
```diff
@@ -50,6 +51,18 @@ def run_query
         expect(results.map(&:id)).to eq(expected_ordered_ids)
       end
 
+      context 'when options contain a pagination_strategy' do
```
ncperry, line 54:
```suggestion
      context 'when options contain a paginator' do
```

spec/brainstem/query_strategies/filter_and_search_spec.rb
```diff
@@ -50,6 +51,18 @@ def run_query
...
+        let(:paginator) { Object.new }
+        let(:options) { default_options.merge(paginator: paginator) }
+
+        it 'uses the provided pagination_strategy' do
```
ncperry, line 58:
```suggestion
        it 'uses the provided paginator' do
```

spec/brainstem/query_strategies/filter_and_search_spec.rb
ncperry: The `pagination_strategy` is something that we moved away from.

spec/brainstem/query_strategies/filter_or_search_spec.rb
```diff
@@ -38,53 +38,21 @@
...
-                to make_database_queries({ count: 1, matching: "SELECT COUNT(distinct `workspaces`.id) FROM" })
-
-              expect { subject.execute(Workspace.unscoped) }.
-                not_to make_database_queries({ count: 1, matching: /SELECT\s+DISTINCT SQL_CALC_FOUND_ROWS workspaces.id FROM/ })
-
-              expect { subject.execute(Workspace.unscoped) }.
-                not_to make_database_queries({ count: 1, matching: "SELECT FOUND_ROWS()" })
-            end
-          end
+        it 'uses the provided pagination_strategy' do
```
ncperry, line 52:
```suggestion
        it 'uses the provided paginator' do
```

PR #160, opened 2021-12-20
Add setting to skip datetime serialization

ethransom:
This happens automatically in rails, probably thanks to
ActiveSupport defining Time.to_json and Date.to_json or
similar. It's a surprisingly large source of CPU burn to do
it here in a pass that rewrites everything.


jarednielson (approved):
LGTM

